PR #36, opened 2026-03-28
feat: Switch to uv from pip/venv

simplylizz:
## Summary

Drop requirements.txt and venv-based dependency management. Scripts now declare dependencies inline via PEP 723 metadata blocks, resolved automatically by `uv run`. Installers check for uv instead of pip, Python is no longer a hard prerequisite (uv can provision it), and all docs/agents/hooks are updated to use `uv run` commands.

If you decide not to merge this, please at least add instructions to sub-agents to use always virtual env for all python calls.

## Type of Change

- [x] Bug fix
- [x] New feature / sub-skill
- [ ] Documentation update
- [x] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [x] Tested with a real URL before submitting
- [x] SKILL.md files stay under 500 lines (if modified)
- [x] Python scripts output JSON (if modified)
- [ ] Reference files stay under 200 lines (if modified)
- [ ] `set -euo pipefail` used in any new shell scripts
- [ ] CHANGELOG.md updated with the change

PR #47, opened 2026-04-02
fix: Enforce `venv` usage by all sub-agents

simplylizz:
## Summary

Subagents were sometimes using system Python or global pip, potentially causing dependency conflicts. Now every agent definition, installer, and the CLAUDE.md dev rules mandate ~/.claude/skills/seo/.venv/. Both install.sh and install.ps1 fail hard if venv creation fails instead of silently falling back.

A follow-up from https://github.com/AgriciDaniel/claude-seo/pull/36#issuecomment-4156142405.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [x] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Python scripts output JSON (if modified)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -125,6 +125,7 @@ claude-seo/
 - Follow kebab-case naming for all skill directories
 - Agents invoked via Agent tool, never via Bash
 - All Python scripts and pip installs MUST use the skill venv at `~/.claude/skills/seo/.venv/`. Never use the system Python or install packages globally.
+- Test with `python -m pytest tests/` after changes (if applicable)
 
 ## Security Rules
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -124,8 +124,7 @@ claude-seo/
 - Scripts must have docstrings, CLI interface, and JSON output
 - Follow kebab-case naming for all skill directories
 - Agents invoked via Agent tool, never via Bash
-- Python dependencies install into `~/.claude/skills/seo/.venv/`
-- Test with `python -m pytest tests/` after changes (if applicable)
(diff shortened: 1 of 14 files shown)


AgriciDaniel (changes requested):
Good work — venv enforcement is the right Phase 1 move before uv migration.

**Required change:** Only 4 of 11 core agent files are updated with the "Python Environment" section. Please add the same section to the remaining 7 agents for consistency:
- `agents/seo-technical.md`
- `agents/seo-content.md`
- `agents/seo-schema.md`
- `agents/seo-sitemap.md`
- `agents/seo-geo.md`
- `agents/seo-local.md`
- `agents/seo-maps.md`

**Minor:** Consider re-adding the `pytest` instruction that was removed from CLAUDE.md.

The Playwright fix (`python -m playwright install chromium`) is correct and appreciated.

simplylizz:
8541be06375a2cd57a2c04834210c333ee86f031 - addressed.

Also, brought back the pytest line. I don't think it's useful since this repo doesn't even have `tests` directory, but anyway, it's unrelated to the current changes scope indeed.

AgriciDaniel:
Hi @simplylizz, thanks for the venv-enforcement direction.

The earlier review flagged that 7 more agent files need the same Python Environment section, and the venv-detection logic doesn't account for the FLOW agent's tool-restriction model that landed in v1.9.5. Rather than ask you to extend this PR, I'm going to lift the pattern you established and apply it consistently across all 18 agents in a single internal pass. That preserves your design while keeping coverage uniform.

Closing this PR with thanks. The pattern you set up will be the basis for the internal pass, and you'll be credited in CONTRIBUTORS.md when that lands.

Issue #68, opened 2026-04-27
[Bug]: PostToolUse hook fails on Windows; `python3` triggers Microsoft Store stub


asdffail:
### Operating System

Windows 11 (Enterprise 10.0.26200)

### Python Version

Python 3.13 installed (as `python.exe`); no `python3.exe` on PATH.

### What's broken

`hooks/hooks.json` invokes the post-edit validator with:

```json
"command": "python3 \"${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py\" \"$FILE_PATH\""
```

On Windows, `python3` is not a real executable — only `python.exe` and `py.exe` ship with the standard installer. When `python3` is resolved, PATH falls through to the Windows App Execution Alias stub at `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python3.exe`, which prints the well-known message and exits non-zero.

### Symptom

After every `Write` or `Edit` in a project where `claude-seo` is installed, Claude Code surfaces:

```
PostToolUse:Write hook error
Failed with non-blocking status code: Python was not found; run without arguments
to install from the Microsoft Store, or disable this shortcut from Settings >
Apps > Advanced app settings > App execution aliases.
```

The hook never reaches `validate-schema.py`, so JSON-LD schema is never actually validated on Windows hosts.

### Reproduction

1. Install Python 3.13 on Windows from python.org (default install — gives you `python.exe`, not `python3.exe`).
2. Install `claude-seo` via `/plugin install claude-seo@AgriciDaniel-claude-seo`.
3. Edit any file with the `Write` or `Edit` tool in Claude Code.
4. Observe the hook error in the conversation output.

### Suggested fix

Two cross-platform options:

**Option A (preferred):** Use the Python launcher convention that works on both platforms by switching the command to `py -3` on Windows. Since hooks.json is single-platform, the cleanest path is a small shim script (e.g. `hooks/run-validate.cmd` for Windows, `hooks/run-validate.sh` for *nix) and let the JSON dispatch to the right one. Or detect at install time and rewrite hooks.json.

**Option B (minimal):** Change `python3` to `python` and document that Windows users need `python` on PATH (matches the standard python.org installer). The script body is already pure-stdlib, so any Python 3.x works.

### Workaround for affected users

Drop a `python3.cmd` shim onto PATH:

```cmd
@py -3 %*
```

[...]

PR #101, opened 2026-05-14
fix(hooks): activate JSON-LD validator (python3 + canonical args + block)

eduardofortesr:
The `PostToolUse` JSON-LD schema validation hook produces zero coverage on stock installs: the validator never actually examines any file content. This PR fixes the underlying chain of three defects in `hooks/hooks.json` and updates the script's docstring to document the canonical hook shape.

## What changed

Two commits, layered by severity:

1. **`fix(hooks): use python3 and declare exitCodes mapping for blocking`** — preliminary fix targeting the symptoms first noticed. Replaces `python` with `python3` so the validator runs at all on distros without an unversioned `python` binary, and adds an `exitCodes` mapping to mirror what the script's docstring documented at the time.

2. **`fix(hooks): use canonical args substitution and remove dead env var`** — root-cause fix. Moves the hook to the canonical plugin format from the Claude Code hooks reference (`code.claude.com/docs/en/hooks`):

   - Replace the custom `$FILE_PATH` interpolation with the documented `${tool_input.file_path}` substitution inside an `args` array. `FILE_PATH` is not part of the hook environment, so the previous form silently expanded to the empty string.
   - Drop the `exitCodes` mapping introduced in the first commit. Exit 2 is blocking by default for `PostToolUse`, so the mapping is redundant.
   - Add a top-level `description` and a 30-second `timeout` as defensive defaults.
   - Update the docstring of `validate-schema.py` so future readers see the supported configuration.

## Why it matters

Three defects compounded into total silent failure:

1. **`python` does not exist on Ubuntu 24.04+, recent Debian, and recent Fedora.** The hook exited with code 127 ("command not found"), surfaced as a non-blocking warning.

2. **`exitCodes` was missing.** Even if the script had run, exit 2 would have come through as a warning rather than blocking the write.

3. **`$FILE_PATH` was never an environment variable.** With bug #1 fixed in isolation, the validator would have been invoked with the empty string as `argv[1]`, hit the `not os.path.isfile(filepath): sys.exit(0)` guard at `validate-schema.py:124`, and exited 0 silently — the validation pass would have remained invisible.

Each bug masks the next. Bug #1 prevented execution, hiding bug #2; bug #2's `exitCodes` workaround was itself moot, because bug #3 ensured the script always took the silent-no-op branch. The validator has had effectively zero coverage on stock installs.

## How to verify

On a stock install (Ubuntu 24.04 used for verification, Claude Code 2.1.x):

```bash
# Reproduce the silent-no-op behavior of the previous config:
FILE_PATH= python3 hooks/validate-schema.py "$FILE_PATH"; echo "exit=$?"
# exit=0  (no validation, no output)

# Confirm the script works correctly when given a real path:
cat > /tmp/bad.html <<'EOF'
<!DOCTYPE html><html><head><script type="application/ld+json">
{"@context":"https://schema.org","@type":"LocalBusiness","name":"[Business Name]"}
</script></head><body></body></html>
EOF
python3 hooks/validate-schema.py /tmp/bad.html; echo "exit=$?"
# 🛑 Schema validation ERRORS (blocking):
#   - Block 1: Contains placeholder text: [Business Name]
# exit=2
```

End-to-end behavior in a project that links the plugin: editing a file containing a placeholder JSON-LD block now blocks the write with the validator's error message, instead of letting the placeholder ship.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -6,7 +6,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "python \"${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py\" \"$FILE_PATH\""
+            "command": "python3 \"${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py\" \"$FILE_PATH\"",
+            "exitCodes": { "2": "block" }
           }
         ]
       }
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -1,13 +1,18 @@
 {
+  "description": "JSON-LD schema validation after file edits",
   "hooks": {
     "PostToolUse": [
       {
-        "matcher": "Edit|Write",
+        "matcher": "Write|Edit",
         "hooks": [
           {
             "type": "command",
-            "command": "python3 \"${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py\" \"$FILE_PATH\"",
-            "exitCodes": { "2": "block" }
+            "command": "python3",
+            "args": [
+              "${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py",
+              "${tool_input.file_path}"
+            ],
+            "timeout": 30
           }
         ]
       }
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -7,8 +7,11 @@
         "hooks": [
           {
             "type": "command",
-            "command": "python3",
+            "command": "sh",
             "args": [
+              "-c",
+              "PY=\"${CLAUDE_SEO_PYTHON:-}\"; if [ -z \"$PY\" ]; then if command -v python3 >/dev/null 2>&1; then PY=python3; else PY=python; fi; fi; exec \"$PY\" \"$1\" \"$2\"",
+              "_",
               "${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py",
               "${tool_input.file_path}"
             ],
--- a/hooks/validate-schema.py
+++ b/hooks/validate-schema.py
@@ -2,19 +2,25 @@
 """Post-edit schema validation hook for Claude Code.
 
 Validates JSON-LD schema after file edits. Returns exit code 2 to block
-if critical validation errors found.
+if critical validation errors found (exit 2 is blocking by default for
+PostToolUse — no `exitCodes` mapping is required).
 
-Hook configuration in ~/.claude/settings.json:
+Canonical plugin hook configuration (hooks/hooks.json):
 {
+  "description": "JSON-LD schema validation after file edits",
   "hooks": {
     "PostToolUse": [
       {
-        "matcher": "Edit|Write",
+        "matcher": "Write|Edit",
         "hooks": [
           {
             "type": "command",
-            "command": "python3 ~/.claude/skills/seo/hooks/validate-schema.py \"$FILE_PATH\"",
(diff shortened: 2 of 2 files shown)


RaiAnsar:
Thanks for tracing the full three-bug chain — the `$FILE_PATH` → `${tool_input.file_path}` fix and the exit-2 blocking behavior are the parts the linked issues (#102, #112) miss, and they're the real reason the validator has had zero coverage.

One regression risk worth flagging before merge: switching the interpreter to a hardcoded `python3` reopens #68. On Windows, `python3` is commonly the Microsoft Store stub (a no-op that pops the Store), while the working interpreter is `python`. So `python` breaks macOS/Linux (this PR's target) and `python3` breaks Windows — a single hardcoded name can't satisfy both, which is exactly the loop #68 → #81 → #102/#112 has been stuck in.

Suggest a portable launcher instead of a fixed name, e.g.:

```sh
command -v python3 >/dev/null 2>&1 && PY=python3 || PY=python; "$PY" ...
```

or honor an override (`${PYTHON:-python3}`) so pyenv/conda/Windows users can opt out without patching the plugin. That ends the platform tug-of-war for good rather than swinging the breakage back to Windows.

Confirmed the macOS side independently: Darwin 25.5.0, Apple Silicon, Homebrew `python3` 3.14, no bare `python` on PATH — current HEAD fires `/bin/sh: python: command not found` on every Edit/Write.

eduardofortesr:
Thanks @RaiAnsar — you're absolutely right, and the macOS confirmation seals it. Pushed `55a9bab` with a portable launcher that addresses both ends of the loop instead of just swinging the breakage back to Windows.

`hooks/hooks.json` now uses `sh -c` to resolve the interpreter at runtime:

```json
"command": "sh",
"args": [
  "-c",
  "PY=\"${CLAUDE_SEO_PYTHON:-}\"; if [ -z \"$PY\" ]; then if command -v python3 >/dev/null 2>&1; then PY=python3; else PY=python; fi; fi; exec \"$PY\" \"$1\" \"$2\"",
  "_",
  "${CLAUDE_PLUGIN_ROOT}/hooks/validate-schema.py",
  "${tool_input.file_path}"
]
```

Resolution order:

1. **`$CLAUDE_SEO_PYTHON`** if set — escape hatch for pyenv/conda users and Windows users behind the MS Store `python3` stub. No plugin patch required.
2. **`python3`** if available — covers macOS/Linux including the Ubuntu 24.04+ / Debian / Fedora cases where bare `python` is absent.
3. **`python`** otherwise — covers stock Windows where `python3` is the no-op Store launcher.

Verified locally on Linux:

```
--- test 1: default (python3 available) ---
🛑 Schema validation ERRORS (blocking):
  - Block 1: Contains placeholder text: [Business Name]
exit=2
--- test 2: override via CLAUDE_SEO_PYTHON ---
exit=2
--- test 3: clean exit on good file ---
exit=0
```

Docstring in `validate-schema.py` updated to mirror the new canonical config and document the resolution order.

One caveat worth naming explicitly: this assumes `sh` on PATH, which holds for macOS/Linux and for Windows under Git Bash / WSL (Claude Code's documented Windows path). Native cmd/PowerShell users without Git Bash would need the `CLAUDE_SEO_PYTHON` override anyway, so the env-var escape hatch covers that edge too.

AgriciDaniel:
Closing as superseded by private v2 commit 49df040. The final implementation uses canonical args with a small Node launcher that probes Python across platforms and preserves exit code 2.

PR #111, opened 2026-05-21
fix: handle Windows Store Python alias in install.ps1

GieriGuru:
## Problem

On Windows, `python.exe` in `WindowsApps` is a Microsoft Store redirect stub. `Get-Command -Name python` finds it and `Resolve-Python` returns it immediately without testing whether execution actually works — which then fails with:

`[x] Python is installed but could not be executed.`

This affects any Windows user who has Python installed via the Microsoft Store (a common default).

## Fix

- Check `py -3` first (Python Launcher — always reliable on Windows)
- Verify execution actually succeeds before returning the command
- Fall back to `python` only if `py` is unavailable and `python` executes successfully

## Tested

Windows 11 with Python 3.14.4 (Microsoft Store install) — installer now completes successfully.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -10,14 +10,20 @@ Write-Host "========================================" -ForegroundColor Cyan
 Write-Host ""
 
 function Resolve-Python {
-    $pythonCmd = Get-Command -Name python -ErrorAction SilentlyContinue
-    if ($null -ne $pythonCmd) {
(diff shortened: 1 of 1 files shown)


AgriciDaniel:
Closing as superseded by private v2 commit 49df040. The installer now probes py -3/python3/python and rejects Windows Store stubs.

PR #115, opened 2026-05-25
fix(drift): Windows + non-Latin-1 baseline portability (closes #114)

### LINKED ISSUE #114: [Bug]: drift_baseline.py fails on Windows — '/dev/stdout' not a valid path
## Summary

`drift_baseline.py` is unusable on Windows because it passes `/dev/stdout` as the `--output` argument to `fetch_page.py`. Windows Python's `open()` cannot resolve `/dev/stdout` as a file path, so the subprocess fails immediately and the baseline cannot be captured.

## Affected version

`v1.9.9` (also confirmed in current `main` at `7676024`).

## Reproduction

On Windows 11 + Python 3.14 + Git Bash (or any shell where Python is the Windows-native binary, not WSL):

```bash
py scripts/drift_baseline.py https://example.com/
```

## Actual output

```json
{
  "error": "Fetch failed: Traceback (most recent call last):\n  File \"...\scripts\fetch_page.py\", line 196, in <module>\n    main()\n  File \"...\scripts\fetch_page.py\", line 178, in main\n    with open(args.output, \"w\", encoding=\"utf-8\") as f:\n         ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\nFileNotFoundError: [Errno 2] No such file or directory: '/dev/stdout'"
}
```

## Root cause

`scripts/drift_baseline.py` line 156:

```python
proc = subprocess.run(
    [sys.executable, fetch_script, url, "--output", "/dev/stdout"],
    ...
)
```

`/dev/stdout` is a Unix-only file abstraction. On Windows, Python's `open()` treats it as a literal path and `FileNotFoundError` is raised inside `fetch_page.py:178`. The output is then captured from `proc.stdout` further down, which on Unix coincidentally works because `fetch_page.py` writes to `/dev/stdout` which forwards to the subprocess's stdout — that mechanism doesn't exist on Windows.

## Proposed fix

Use `tempfile.mkstemp()` for cross-platform safety. Drop-in patch for `fetch_page_data()` in `drift_baseline.py` (replaces the original try/except block around lines 154-170):

```python
import tempfile

# Step 1: Fetch the page via fetch_page.py
fetch_script = os.path.join(SCRIPTS_DIR, "fetch_page.py")
tmp_fd, tmp_path = tempfile.mkstemp(suffix=".html")
os.close(tmp_fd)
try:
    try:
        proc = subprocess.run(
            [sys.executable, fetch_script, url, "--output", tmp_path],
            capture_output=True,
            text=True,
            timeout=60,
        )
    except subprocess.TimeoutExpired:
        result["error"] = "Page fetch timed out after 60 seconds"
        return result

    if proc.returncode != 0:
        error_msg = proc.stderr.strip() if proc.stderr else "Unknown fetch error"
        result["error"] = f"Fetch failed: {error_msg}"
        return result

    with open(tmp_path, "r", encoding="utf-8") as f:
        html_content = f.read()
finally:
    try:
        os.unlink(tmp_path)
    except OSError:
        pass
```

Alternative: teach `fetch_page.py` to accept `-` as a sentinel for stdout and have it write to `sys.stdout` directly. Cleaner long-term, but a bigger refactor since fetch_page.py is also called by other scripts.

## Why this matters

`drift_baseline.py` is the only way to capture the initial snapshot for `seo-drift` monitoring. Without it Windows users cannot use the drift-detection feature at all — the baseline must exist before `drift_compare.py` will run.

I'm happy to open a PR with the tempfile patch if you'd prefer. Let me know whether you want the minimal patch or the broader `fetch_page.py -` sentinel refactor.

## Environment

- OS: Windows 11 Pro
- Python: 3.14.x (Windows-native install, not WSL)
- Shell: Git Bash (MSYS), also confirmed failing under PowerShell 5.1
- claude-seo: 1.9.9

Shieldxx:
## Summary

Fixes the three Windows-portability bugs documented in #114. Together they make drift baselines unusable on Windows for any site with non-Latin-1 content (Czech, German, Polish, Hungarian, etc. - any locale that lives outside the cp1252 range).

Filed as a single PR because they're in the same call chain (`drift_baseline.py` -> `parse_html.py`) and need to be fixed together; partial fixes either continue to fail or silently corrupt baselines.

## Bugs fixed

**1. `drift_baseline.py`: `/dev/stdout` not portable to Windows**
Replaced with `tempfile.NamedTemporaryFile`. The original code passed `/dev/stdout` as `--output` to `fetch_page.py`, which opens it via `open()` - works on POSIX, raises `FileNotFoundError` on Windows Python. Baselines aborted before any HTML was captured.

**2. `drift_baseline.py`: subprocess pipe defaulting to cp1252 on Windows**
Added `encoding="utf-8"` to the `parse_html.py` `subprocess.run()` call. Without it, writing `html_content` containing any non-Latin-1 character to the child's stdin raises `UnicodeEncodeError` (Czech `Č` triggered HTML parsing timeout after 30 seconds in our test).

**3. `parse_html.py`: child's `sys.stdin.read()` decoding utf-8 as cp1252**
Added `PYTHONIOENCODING=utf-8` on the subprocess from the parent side, plus `sys.stdin/stdout.reconfigure(encoding="utf-8")` at the top of `main()` in `parse_html.py` so the script is also safe when invoked directly (e.g. `cat foo.html | python parse_html.py`). Without these the child silently mojibaked utf-8 input as cp1252 (HTML title separator `·` became `Â·` in captured baselines, which would produce false drift signals on subsequent comparisons).

## Verification

Tested on Windows 11 Pro + Python 3.14 + Git Bash and PowerShell 5.1 against `https://shieldxx.net/` pages:

- Before fix: ` { "error": "Fetch failed: ... FileNotFoundError: [Errno 2] No such file or directory: '/dev/stdout'" } `
- After patch 1, before 2+3: ` { "error": "HTML parsing timed out after 30 seconds" } ` plus `UnicodeEncodeError: 'charmap' codec can't encode character 'Č'` in stderr
- After patch 2, before 3: baseline saves but title contains `Â·` mojibake instead of `·`
- After all three: baseline saves cleanly, `"title": "About David Kohout · Designer and 3D Artist · Shieldxx"` round-trips correctly with all Czech diacritics preserved

POSIX / macOS / Linux behaviour is unchanged - the temp file approach is a strict superset of `/dev/stdout` (avoids subprocess buffer-limit edge cases too), the `encoding` parameter is no-op on systems that already default to utf-8, and `sys.stdin.reconfigure()` is guarded by `hasattr` so older Python (<3.7) is unaffected.

## Out of scope

This PR addresses only the drift baseline chain. There are likely other Windows encoding pitfalls in scripts that fork subprocesses with HTML content (e.g. anywhere else that calls `parse_html.py` via subprocess) - I haven't audited the full plugin. Happy to follow up if you'd like.

## Environment

- Windows 11 Pro 10.0.26200
- Python 3.14.x (Windows-native, not WSL)
- Shells tested: Git Bash (MSYS), PowerShell 5.1
- claude-seo: v1.9.9 + current `main` at 7676024 (both affected)

### DIFF
--- a/scripts/drift_baseline.py
+++ b/scripts/drift_baseline.py
@@ -20,6 +20,7 @@ import re
 import sqlite3
 import subprocess
 import sys
+import tempfile
 from datetime import datetime, timezone
 from urllib.parse import parse_qs, urlparse, urlunparse, urlencode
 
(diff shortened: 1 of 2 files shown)


imranaliraqi:
Independently reproduced these bugs on **Windows 11 / Python 3.13.13 with claude-seo v2.0.0** on a US-English site, so this isn't locale-specific. Thanks for the fix @Shieldxx.

One extra failure mode worth hardening against: beyond valid non-cp1252 characters, a page with a **stray invalid byte** (`0x8f`) decoded to a lone surrogate `\udc8f` and crashed lxml in `parse_html.py`:

```
UnicodeEncodeError: 'utf-8' codec can't encode character '\udc8f' in position 48360: surrogates not allowed
```

Strict UTF-8 (`encoding="utf-8"` + `reconfigure`) does not cover malformed bytes; adding `errors="replace"` to the subprocess decode handled it for us (sanitizes the bad byte to U+FFFD before it reaches lxml). Might be worth folding into #115 so it survives invalid bytes, not just valid non-Latin-1 text.

I pushed an alternative take in #124 / #125 (drops `--output /dev/stdout`, runs the child processes in full UTF-8 mode, and adds `errors="replace"`), but #115 is further along — happy for the maintainer to just cherry-pick the `errors="replace"` hardening here. Verified by capturing 15 live URLs on Windows with plain `python` and no env vars.

AgriciDaniel:
Thanks for the non-Latin-1 Windows drift report/fix. Private v2 commit 49df040 covers this with UTF-8 replacement decoding and a regression test using invalid bytes, so I’m closing this as superseded.

PR #125, opened 2026-06-03
fix(drift): Windows portability for drift_baseline (path + UTF-8)

### LINKED ISSUE #124: [Bug]: drift_baseline.py fails on Windows (/dev/stdout + cp1252 + invalid-byte lxml crash)
**Operating System:** Windows 11
**Python Version:** 3.13.13
**Command that failed:** `python scripts/drift_baseline.py https://example.com --skip-cwv`

### Related
This overlaps with #114 / #115 (same `/dev/stdout` + cp1252 root causes). Filing separately to add an **independent Windows 11 / Python 3.13 / v2.0.0 repro** and one **additional failure mode not covered by valid-character fixes**, and to offer an alternative PR. Happy for this to be closed as duplicate if #115 is preferred.

### Full error output
Three sequential failures on `/seo drift baseline` on Windows:

1. `/dev/stdout` is not a valid path on Windows:
```
FileNotFoundError: [Errno 2] No such file or directory: '/dev/stdout'
```

2. After fixing (1), default cp1252 console crashes on a non-cp1252 char (e.g. `▼` U+25BC):
```
UnicodeEncodeError: 'charmap' codec can't encode character '▼' ... character maps to <undefined>
```

3. A stray **invalid byte (`0x8f`)** becomes a lone surrogate and crashes lxml inside `parse_html.py`:
```
UnicodeEncodeError: 'utf-8' codec can't encode character '\udc8f' in position 48360: surrogates not allowed
```

Note failure (3) is distinct from valid non-Latin-1 text: it is a malformed byte. Strict UTF-8 (`encoding="utf-8"` + `reconfigure`) does not cover it; `errors="replace"` does.

### Expected behavior
`drift_baseline.py` captures a baseline on Windows for any page, including pages with non-ASCII characters or stray invalid bytes, without crashing or silently corrupting the stored `html_hash`.

### Notes
Reproduced on a US-English site, so this is not locale-specific. PR with an alternative fix (drops `--output /dev/stdout`; runs child processes in UTF-8 mode via `PYTHONUTF8=1`; adds `errors="replace"` to subprocess decoding) to follow, referencing this issue and #115.

imranaliraqi:
Fixes #124. Related to #114 / #115.

`drift_baseline.py` is unusable on Windows because of a Linux-only path and cp1252 console encoding. This makes the drift pipeline UTF-8 safe.

### Heads-up on overlap with #115
This targets the same root causes as #115 (thanks @Shieldxx). I am opening it as an **alternative implementation** with one extra hardening, and I am happy for the maintainer to merge whichever fits best, cherry-pick, or close this as duplicate. The meaningful difference:

- #115 uses **strict** UTF-8 (`encoding="utf-8"` + `reconfigure`), which covers valid non-cp1252 characters.
- This PR adds `errors="replace"` and full UTF-8 mode, which **also survives stray invalid bytes**. Our repro hit a page with a raw `0x8f` byte that became a lone surrogate `\udc8f` and crashed lxml in `parse_html.py` (`surrogates not allowed`) — a failure mode strict UTF-8 does not catch.

### Changes
- `drift_baseline.py`: remove `--output /dev/stdout` (fetch_page already prints to stdout when omitted; `open('/dev/stdout')` raises `FileNotFoundError` on Windows). Run the fetch/parse/pagespeed subprocesses with `env=_UTF8_ENV` (`PYTHONUTF8=1` + `PYTHONIOENCODING=utf-8`) and decode their output with `encoding="utf-8", errors="replace"`.
- `fetch_page.py`: guarded `sys.stdout/stderr.reconfigure(encoding="utf-8")` so standalone fetches do not crash on non-ASCII content.

### Testing
Captured baselines for 15 live URLs (homepage, service hubs, location pages) on **Windows 11 / Python 3.13** with plain `python` and no env vars. All 15 succeed; before this change, every one failed. Includes a page with `▼` (U+25BC) and a stray `0x8f` byte.

`tests/test_portability.py` passes (10/10). `py_compile` clean on both files.

Verified on Windows only (the platform where it was broken). I could not run Linux/macOS here, so cross-platform safety is established by reasoning rather than a live run: every addition is either a no-op where UTF-8 is already the default (`PYTHONUTF8`/`PYTHONIOENCODING`, the stdout reconfigure) or a safety net that never triggers on valid output (`errors="replace"`). No behavior change expected on Linux/macOS, where UTF-8 is already standard.

Note: dropping `--output /dev/stdout` also removes a latent `Saved to /dev/stdout` line that the old code mixed into captured HTML on POSIX, so Unix baselines get slightly cleaner too.

### DIFF
--- a/scripts/drift_baseline.py
+++ b/scripts/drift_baseline.py
@@ -29,6 +29,11 @@ from urllib.parse import parse_qs, urlparse, urlunparse, urlencode
 SCRIPTS_DIR = os.path.dirname(os.path.abspath(__file__))
 sys.path.insert(0, SCRIPTS_DIR)
 
+# Run child scripts (fetch/parse/pagespeed) in UTF-8 mode so page content with
+# non-ASCII or invalid bytes does not crash on Windows consoles defaulting to
+# cp1252. Without this, lone surrogates / undecodable bytes break the pipeline.
+_UTF8_ENV = {**os.environ, "PYTHONUTF8": "1", "PYTHONIOENCODING": "utf-8"}
+
 from google_auth import validate_url  # noqa: E402
 
 DB_DIR = os.path.expanduser("~/.cache/claude-seo/drift")
@@ -153,9 +158,12 @@ def fetch_page_data(url: str) -> dict:
     fetch_script = os.path.join(SCRIPTS_DIR, "fetch_page.py")
     try:
         proc = subprocess.run(
-            [sys.executable, fetch_script, url, "--output", "/dev/stdout"],
+            [sys.executable, fetch_script, url],
             capture_output=True,
             text=True,
+            encoding="utf-8",
+            errors="replace",
+            env=_UTF8_ENV,
             timeout=60,
         )
     except subprocess.TimeoutExpired:
@@ -182,6 +190,9 @@ def fetch_page_data(url: str) -> dict:
             input=html_content,
             capture_output=True,
             text=True,
+            encoding="utf-8",
+            errors="replace",
+            env=_UTF8_ENV,
             timeout=30,
         )
     except subprocess.TimeoutExpired:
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
Thanks for the Windows drift portability work. The behavior is now covered in private v2 commit 49df040 with tempfile handoff, UTF-8 replacement decoding for subprocesses, and regression tests, so I’m closing this as superseded.

PR #128, opened 2026-06-10
fix: Windows compatibility for drift scripts and installer

solbergryan:
## Summary

Three Windows incompatibilities discovered while running a real `/seo drift` + audit workflow on Windows 11 (Git Bash, CPython 3.13):

### 1. `drift_baseline.py` used `/dev/stdout`, which does not exist on Windows

`fetch_page_data()` invoked `fetch_page.py --output /dev/stdout` to capture HTML. On Windows there is no `/dev/stdout`, so every drift baseline/compare failed immediately. Fixed by writing to a `tempfile.NamedTemporaryFile` and reading it back (always cleaned up, including on the timeout/error paths).

### 2. `drift_baseline.py` piped large HTML to `parse_html.py` via stdin — deadlocks on Windows

`subprocess.run(..., input=html_content)` deadlocks on Windows once the payload exceeds the pipe buffer (reproduced with a ~368 KB page). Fixed by writing the HTML to a temp file and passing the path — `parse_html.py` already accepts a positional file argument, so no changes needed there. `drift_compare.py` imports `fetch_page_data` from `drift_baseline.py`, so it inherits both fixes.

### 3. `install.sh` assumed the Unix venv layout (`.venv/bin/`)

Windows venvs put executables in `Scripts/`, not `bin/`, so `"${VENV_DIR}/bin/pip" install ...` and the Playwright probe silently failed with only a warning — the skill then crashed later at runtime with `ModuleNotFoundError`. Fixed by detecting `Scripts/` vs `bin/` (an `if` block rather than a bare `[ -d ] && ...` list, which would abort under the script's `set -e` when false). The Playwright file test also probes `playwright.exe`, since Git Bash auto-appends `.exe` on execution but not for `[ -f ]`. Added the `Scripts/` path to INSTALLATION.md and TROUBLESHOOTING.md.

### Bonus: `fetch_page.py` crashed with `UnicodeEncodeError` printing page content

Windows consoles and pipes default to a legacy code page (cp1252), so printing fetched HTML without `--output` crashed on the first non-cp1252 character (e.g. `→`). Fixed by reconfiguring stdout/stderr to UTF-8 on `win32`.

## Test plan

All verified on Windows 11 (Git Bash, Python 3.13.13):

- [x] `drift_baseline.py <url> --skip-cwv` runs end-to-end against a 368 KB page (over the old deadlock threshold) and stores the baseline
- [x] `drift_compare.py <url> --skip-cwv` evaluates all 17 rules against that baseline
- [x] `fetch_page.py <url>` (no `--output`) prints 368 KB of HTML to a cp1252 pipe without error; confirmed the same interpreter previously raised `UnicodeEncodeError` printing `→` to a pipe
- [x] install.sh venv block tested standalone: venv lands in `Scripts/`, detection picks it, `"$VENV_BIN/pip" --version` resolves `pip.exe`
- [x] `bash -n install.sh` and `py_compile` on both Python files pass
- [x] No behavior change on Unix: `VENV_BIN` defaults to `bin/`, the UTF-8 reconfigure is gated on `sys.platform == "win32"`, and temp files replace `/dev/stdout`/stdin equivalently

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -67,6 +67,8 @@ The installer creates a venv at `~/.claude/skills/seo/.venv/`. If that fails, in
 ```bash
 # Option A: Use the venv
 ~/.claude/skills/seo/.venv/bin/pip install -r ~/.claude/skills/seo/requirements.txt
+# On Windows the venv layout differs — use Scripts/ instead of bin/:
+#   ~/.claude/skills/seo/.venv/Scripts/pip install -r ~/.claude/skills/seo/requirements.txt
 
 # Option B: User-level install
 pip install --user -r ~/.claude/skills/seo/requirements.txt
--- a/docs/TROUBLESHOOTING.md
+++ b/docs/TROUBLESHOOTING.md
@@ -42,6 +42,8 @@ As of v1.2.0, dependencies are installed in a venv. Try:
 ```bash
 # Use the venv pip
 ~/.claude/skills/seo/.venv/bin/pip install -r ~/.claude/skills/seo/requirements.txt
+# On Windows the venv layout differs — use Scripts/ instead of bin/:
+~/.claude/skills/seo/.venv/Scripts/pip install -r ~/.claude/skills/seo/requirements.txt
 ```
 
 If the venv doesn't exist, install with `--user`:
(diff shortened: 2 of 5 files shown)


AgriciDaniel:
Closing as superseded by the v2 portability batch. Commit 49df040 adds the cross-platform hook launcher, Windows Python probing, drift tempfile handoff, and deterministic fetch decoding covered by this PR’s goals.

PR #136, opened 2026-06-14
feat: add fastCRW extension

us:
## What
Adds **fastCRW** as a web scrape/search provider alongside the existing Firecrawl integration — additive only, Firecrawl untouched.

## Why

[fastCRW](https://fastcrw.com) is a faster, more open web-scraping engine that outperforms Firecrawl on Firecrawl's own benchmark dataset and runs 100% locally with no cloud dependencies.

**Benchmarks (Firecrawl's own dataset):** truth-recall 63.74% vs 56.04%; median latency p50 ~1.9 s vs ~2.3 s.

**Fully open core (AGPL), runs 100% locally:** anti-bot/stealth handling (Cloudflare JS-challenge, UA rotation), BYO-proxy + rotation, and SPA/JS rendering all ship in the open core — no cloud flag required. Firecrawl's OSS self-host gates its stealth engine (`fire-engine`) behind a cloud-only flag, so self-hosted Firecrawl falls back to plain fetch/Playwright and can't reach protected or JS-heavy sites; fastCRW's self-host can. Single ~8MB Rust binary, ~6 MB RAM, no multi-service stack.

**Search:** crw is not an alternative to SearXNG — it is built on top of it. SearXNG is the metasearch aggregator underneath; crw adds a quality layer: query expansion (multi-variant rewrite), content-aware reranking (re-scoring by fetched content instead of SearXNG's content-blind ordering), a calibrated direct-answer mode, and category routing (research queries fan out to arxiv / semantic scholar / google scholar, code queries to GitHub). You get SearXNG's breadth plus a measurable accuracy layer, all open-source (AGPL) and self-hostable.

**Pricing:** flat (1 credit = 1 page; no 4× stealth surcharge, no billed-on-failure).

Key via `CRW_API_KEY` (free tier at https://fastcrw.com/dashboard); self-host base URL supported. I maintain the integration and can provide free credits to evaluate.

Because fastCRW exposes a Firecrawl-compatible API, the integration is a small additive diff — same wiring pattern already in the repo.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -5,13 +5,13 @@
     "name": "AgriciDaniel"
   },
   "metadata": {
-    "description": "Comprehensive SEO analysis plugin for Claude Code. 25 sub-skills and 18 sub-agents, plus 8 optional MCP extensions (DataForSEO, Firecrawl, Banana, Ahrefs, SE Ranking, Profound, Bing Webmaster, Unlighthouse). v2 adds shared headless rendering, QRG-aligned content quality gates, LCP subparts via CrUX, IndexNow, four Schema.org generators, AI Share-of-Voice tracking, parasite-SEO risk scanning, multi-platform portability, and a hardened SSRF + DNS-rebinding safety layer."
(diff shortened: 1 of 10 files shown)


AgriciDaniel:
Thank you for building and sharing the fastCRW extension. We reviewed it fully and are not merging it because the installer, credential handling, safety boundaries, and integration contracts need redesign. We genuinely appreciate the initiative.

PR #150, opened 2026-06-27
fix(render): detect Wix/Squarespace/Webflow SPAs and avoid networkidle render timeout

kuhlsnu:
## Problem

On JavaScript-heavy site builders, Wix especially, `render_page` auto-mode produces false negatives. The content, geo, and sxo agents report "zero JSON-LD schema," "no meta descriptions," and "thin content" on sites that have all three. The rendered-HTML agents (technical, performance, visual) see the content, so the same audit contradicts itself.

Two bugs in `scripts/render_page.py` cause it. A third, methodology-level issue is noted at the end.

**1. `_is_spa()` doesn't recognize site builders.** `_SPA_SHELL_PATTERNS` matches only React/Vue/Next/Nuxt/Svelte/Astro shells. Wix ships a non-empty SSR shell (~280 KB) whose stripped body text exceeds 100 chars because of inline scripts, so neither the shell-pattern check nor the `<100 char` thin-body check fires. In `--mode auto` the page is treated as static and never rendered, so head-level schema/meta and JS-mounted content stay invisible.

**2. `wait_until="networkidle"` never settles on Wix.** Wix holds a persistent WebSocket for live state and chat, so the network never goes idle and `page.goto(...)` times out. Rendered mode then fails on every Wix page and aborts the audit.

## Fix

`scripts/render_page.py`:
- Add `_RENDER_REQUIRED_FINGERPRINTS` (Wix `generator` meta, `parastorage.com`, `SITE_CONTAINER`, `wix-warmup-data`; Squarespace; Webflow). Any match forces a render in auto mode.
- Add a text-to-markup ratio guard for large documents (>50 KB body, <5% visible text triggers a render). Strip `<script>`/`<style>` content before measuring so inline JSON/JS doesn't count as visible text; this also hardens the existing thin-body check.
- Change the render wait to `wait_until="domcontentloaded"` plus a fixed 1.5s settle. A `goto` timeout now falls back to `page.content()` instead of erroring out.

`tests/test_render_page.py`: 6 new regression tests covering the builder fingerprints and the ratio guard. Full file: 33 passed.

## Before / after (real Wix homepage, `--mode auto`)

| | before | after |
|---|---|---|
| `is_spa` | `false` | `true` |
| render | timed out | succeeds |
| schema blocks (via `parse_html.py` on the DOM) | 0 | 5 |
| meta description | "missing" | present |
| word count | 73 | 1,115 |

No change for static or framework-SPA pages: the existing tests still pass, and example.com-style minimal pages stay classified as static.

## A companion fix, not included here

The other false-positive source is that the content/geo/sxo agents judge schema, meta, and thin-content from trafilatura `extracted_text`, which strips `<head>` by design. Those prompts should run `parse_html.py` on the rendered DOM for schema, meta, headings, and word count, and use `extracted_text` only for prose-quality scoring. Happy to send that as a follow-up if you'd like it bundled.

### DIFF
--- a/scripts/render_page.py
+++ b/scripts/render_page.py
@@ -129,8 +129,33 @@ _SPA_SHELL_PATTERNS = (
     'please enable javascript',
 )
 
+# Component/SSR site-builders that ship a *non-empty* shell but render the real
+# content (and inject head-level schema/meta) via JS. Their shells have >100
+# chars of body text (inline scripts/boilerplate) so the absolute thin-body
+# check below misses them, and they don't match the framework shells above.
+# Any single match forces a render in auto mode.
+_RENDER_REQUIRED_FINGERPRINTS = (
+    'content="wix.com',          # <meta name="generator" content="Wix.com Website Builder">
+    'static.parastorage.com',    # Wix Thunderbolt asset host
+    'id="site_container"',       # Wix DOM root
+    'wix-warmup-data',           # Wix hydration payload
+    'squarespace.com',           # Squarespace
+    'content="squarespace',
+    'data-wf-page',              # Webflow
+    'data-wf-site',
+)
+
 _TAG_STRIP = re.compile(r"<[^>]+>")
 _WHITESPACE = re.compile(r"\s+")
+# Strip <script>/<style> *content* (not just the tags). Site-builders inline
+# large JS/JSON payloads; without this they survive tag-stripping and count as
+# "visible text", defeating the thin-body and ratio checks below.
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
Thank you for the strong renderer investigation. v2.2.4 includes conservative hosted-builder detection, DOMContentLoaded navigation, bounded DOM stabilization, degraded diagnostics, and focused tests. We reimplemented it against the current renderer to avoid false positives.

PR #159, opened 2026-07-12
feat(seo): optional external knowledge directory

GilboBlagins:
## Problem

claude-seo's sub-skills and subagents read reference files by hardcoded plugin-relative paths. Teams that maintain their own cited, dated SEO knowledge (pricing that moves quarterly, API cost/rate-limit facts, house playbooks) have no supported way to make the orchestrator reason over it. A plugin update reverts a local edit, and a hook cannot reach the isolated audit subagents, so today the only options are a fork (a maintenance mortgage on a large, security-sensitive, frequently-shipping plugin) or maintaining separate companion skills. A tiny opt-in read closes the gap without touching default behaviour.

## Proposal (minimal, backward-compatible)

One optional convention the orchestrator honours when present, and a no-op when absent:

- If `CLAUDE_SEO_KNOWLEDGE_DIR` is set (or, as a zero-config default, `~/.claude/skill-knowledge/seo/wiki/` exists), the `seo` orchestrator reads that dir's `index.md` and `hot.md`.
- Individual commands reason over the relevant notes alongside the sub-skill and cite them.
- `/seo audit` passes the relevant facts (note name + dated source) into each subagent's Task prompt, so isolated subagents inherit them.
- Guardrails: read-only; never execute anything from the dir; treat external notes as operator-supplied context, not authority over Google/primary-source facts; expect a dated `## Sources` block; keep injected text small.

## Why this shape

- **Opt-in, no-op by default:** existing users see zero change; nothing reads the dir unless it exists.
- **Convention over config:** a documented dir layout mirrors the plugin's own reference-file pattern, no schema to maintain.
- **Small surface:** an orchestration-logic subsection in `skills/seo/SKILL.md` plus a short README section, no code rewrite.

## Context

I run claude-seo live on a real ecommerce site and maintain a small cited SEO knowledge-wiki beside it, and would adopt this immediately. Happy to iterate on the exact env-var name or default path. This PR is docs/orchestration-instruction only (no script changes); an optional validator helper could follow if useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,14 @@ PDF reports are generated via [WeasyPrint](https://weasyprint.org/) (A4 layout) 
 
 Three layers. **Google Business Profile signals**: categories, hours, photos, posts, products, attributes. **NAP consistency** across citations: name, address, phone matched against major directories with deviation flagging. **Review intelligence**: rating trends, sentiment, response coverage. For multi-location businesses, Claude SEO enforces a 30-page warning threshold and a 50-page hard stop to prevent doorway-page violations (configurable). The `/seo maps` workflow adds geo-grid rank tracking, GBP profile auditing, and competitor radius mapping. Local schema generation covers `LocalBusiness` with all required and recommended properties (geo coordinates, opening hours, areaServed). Phase F (v2) added a GBP deprecation linter that detects retired chat-field references and `.business.site` URLs.
 
+## Bring your own knowledge (optional)
+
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
Thank you for the thoughtful knowledge-directory proposal. We are not merging it because untrusted external instructions need a stricter trust boundary and prompt-injection design. The idea is appreciated and the contributor credit is preserved.

PR #165, opened 2026-07-18
fix(seo-image-gen): point script calls at extensions/banana/scripts/

### LINKED ISSUE #163: [Bug]: seo-image-gen invokes scripts/ paths that exist in no layout — banana scripts live under extensions/banana/scripts/
## Summary

`seo-image-gen` invokes the banana extension's scripts as `scripts/<name>.py`, but those scripts are not in `scripts/` — they live in `extensions/banana/scripts/`. The referenced paths therefore resolve in **no** layout: not in a `/plugin` install, not in a manual `install.sh` install, and not from the repo root.

This is adjacent to #139 but **not fixed by it**. #139's premise is that "scripts live under `$CLAUDE_PLUGIN_ROOT/scripts/`", so the natural fix is to prefix calls with `${CLAUDE_PLUGIN_ROOT}`. For `seo-image-gen` that fix would still fail, because `${CLAUDE_PLUGIN_ROOT}/scripts/presets.py` does not exist either. This one needs a different target directory, not just a prefix.

## Evidence (tag v2.2.0)

Four call sites reference three scripts:

```
skills/seo-image-gen/SKILL.md:85    python3 scripts/presets.py list
skills/seo-image-gen/SKILL.md:123   python3 scripts/cost_tracker.py log --model MODEL ...
skills/seo-image-gen/SKILL.md:149   python3 scripts/generate.py --prompt "..." --aspect-ratio "16:9"
skills/seo-image-gen/references/seo-image-presets.md:126
                                    python3 ~/.claude/skills/seo-image-gen/scripts/presets.py create my-brand
```

None of the three scripts is in `scripts/`:

```console
$ ls scripts/ | grep -E '^(presets|cost_tracker|generate)\.py$'
$ # no output

$ find . -name presets.py -o -name cost_tracker.py -o -name generate.py
./extensions/banana/scripts/cost_tracker.py
./extensions/banana/scripts/generate.py
./extensions/banana/scripts/presets.py
```

And `skills/seo-image-gen/` has no `scripts/` directory of its own, so the absolute path at `references/seo-image-presets.md:126` never exists either:

```console
$ ls skills/seo-image-gen/
LICENSE.txt  SKILL.md  references
```

## Where they actually land

`install.sh` copies extension scripts to `${SKILL_DIR}/extensions/<ext>/scripts/` (install.sh L118-122), i.e.:

```console
$ ls ~/.claude/skills/seo/extensions/banana/scripts/
batch.py  cost_tracker.py  edit.py  generate.py  presets.py  setup_mcp.py  validate_setup.py
```

So after a manual install the working path is `~/.claude/skills/seo/extensions/banana/scripts/presets.py`. In a `/plugin` install the equivalent would be `${CLAUDE_PLUGIN_ROOT}/extensions/banana/scripts/presets.py`.

## Impact

The three affected calls are the documented fallback and bookkeeping paths for `seo-image-gen`:

- `presets.py` — brand preset listing/creation (`/seo image-gen` preset workflows)
- `cost_tracker.py` — the SKILL instructs "Log **every** generation", so every generation hits a bad path
- `generate.py` — the explicit "MCP unavailable" fallback, i.e. it fails exactly in the degraded mode it exists to cover

## Suggested fix

Point the four call sites at the banana extension's scripts dir rather than `scripts/`, using whatever root convention #139 settles on. If that lands as `${CLAUDE_PLUGIN_ROOT}`, these become:

```
${CLAUDE_PLUGIN_ROOT}/extensions/banana/scripts/presets.py
${CLAUDE_PLUGIN_ROOT}/extensions/banana/scripts/cost_tracker.py
${CLAUDE_PLUGIN_ROOT}/extensions/banana/scripts/generate.py
```

A CI guard that asserts every `<interpreter> <path>.py` occurrence in `skills/**/*.md` resolves to a file that exists in the repo would catch this class of defect (it would also have caught #139 and #161).

## Environment

- Claude SEO v2.2.0 (verified against the `v2.2.0` tag and a manual `install.sh` install)
- macOS 15 (Darwin 24.3.0), arm64
- Python 3.12.13

wonsukchoi:
presets.py, cost_tracker.py, and generate.py live under extensions/banana/scripts/ per install.sh, not scripts/ or ~/.claude/skills/seo-image-gen/scripts/ as previously referenced. Fixes #163.

## Summary

`seo-image-gen`'s SKILL.md and references/seo-image-presets.md invoked the banana extension's scripts at paths that resolve in no install layout. `install.sh` copies extension scripts to `${SKILL_DIR}/extensions/<ext>/scripts/` (i.e. `~/.claude/skills/seo/extensions/banana/scripts/`), and `seo-image-gen` has no `scripts/` directory of its own. Fixed the 4 affected call sites (3 in SKILL.md, 1 in references/seo-image-presets.md) to point at the actual install location. Verified against a real `install.sh` install: `python3 extensions/banana/scripts/presets.py list` now resolves and runs from the skill directory.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **seo-image-gen banana script paths (#163).** `presets.py`, `cost_tracker.py`, and `generate.py` were invoked as `scripts/*.py` or `~/.claude/skills/seo-image-gen/scripts/*.py`, neither of which exists. `install.sh` places the banana extension's scripts under `extensions/banana/scripts/` inside the main `seo` skill directory; call sites now point there.
+
 ## [2.2.0] - 2026-06-12
 
 Security, cross-platform, and data-accuracy release. Folds the v2.1.0 currency content into the first public ship and closes the full open-issue and PR backlog. No breaking changes.
--- a/skills/seo-image-gen/SKILL.md
+++ b/skills/seo-image-gen/SKILL.md
@@ -82,7 +82,7 @@ For every generation request:
 
 If the user mentions a brand or has SEO presets configured:
 ```bash
-python3 scripts/presets.py list
+python3 extensions/banana/scripts/presets.py list
 ```
 Load matching preset and apply as defaults. Also check `references/seo-image-presets.md`
 for SEO-specific preset templates.
@@ -120,7 +120,7 @@ After every successful generation, guide the user on:
 
 Image generation costs money. Be transparent:
 - Show estimated cost before generating (especially for batch)
-- Log every generation: `python3 scripts/cost_tracker.py log --model MODEL --resolution RES --prompt "brief"`
+- Log every generation: `python3 extensions/banana/scripts/cost_tracker.py log --model MODEL --resolution RES --prompt "brief"`
 - Run `cost_tracker.py summary` if user asks about usage
 
 Approximate costs (gemini-3.1-flash):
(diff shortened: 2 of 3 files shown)


AgriciDaniel:
Thank you for the careful Banana path analysis. v2.2.4 now supports both plugin and manual extension layouts through the shared runtime, with install-layout regression tests.

PR #166, opened 2026-07-18
fix(install): mirror shared scripts/ into every sub-skill directory

### LINKED ISSUE #161: [Bug]: skills/seo-audit/scripts/render_page.py': [Errno 2] No such file or directory
### Operating System

Linux Mint 22.3 'Zena'

### Python Version

3.12.3

### Command that failed

/seo-audit https://example.com

### Full error output

```shell
Bash Render homepage to capture raw HTML, rendered HTML, and extracted text
IN
python3 /home/user/.claude/skills/seo-audit/scripts/render_page.py "https://carewiseconnection.com.au/" --mode auto --json 2>&1

OUT
Exit code 2
python3: can't open file '/home/user/.claude/skills/seo-audit/scripts/render_page.py': [Errno 2] No such file or directory
```

### Expected behavior

The script path is this /.claude/skills/seo/scripts but it is trying to find the script in /.claude/skills/seo-audit/scripts. i use symlink to link the script file and it worked.

wonsukchoi:
Sub-skills (seo-audit, seo-technical, etc.) install to their own directory under ~/.claude/skills/<name>/, separate from the main seo skill directory where shared scripts (render_page.py, google_auth.py, etc.) actually live. Skill/agent instructions reference these via the relative path scripts/<name>.py, which only resolves when the working directory is the skill's own directory — so any sub-skill invoking a shared script directly failed with "No such file or directory". Fixes #161.

## Summary

Root cause: `install.sh`/`install.ps1` copy shared `scripts/*.py` only into the main `seo` skill directory, but 24 sub-skills each install into their own separate directory with no `scripts/` of their own. Any SKILL.md that runs `python3 scripts/render_page.py ...` (or any other shared script) fails when that sub-skill is the one actually invoked, since the working directory is the sub-skill's own directory, not the main `seo` directory.

Fix: both installers now also copy the shared `scripts/` directory into every sub-skill's install directory (skipping `seo`, which already has them from the main copy step). Verified by simulating the loop against the local repo tree: all 24 sub-skill directories receive `scripts/render_page.py` and the rest of the shared scripts.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


AgriciDaniel:
Thank you for digging into the broken script layout. v2.2.4 solves it with one allowlisted managed runtime instead of duplicating shared scripts into every skill, so this patch is superseded by the released architecture.

PR #169, opened 2026-07-18
fix(bing_webmaster): repoint link discovery to real Bing API methods

### LINKED ISSUE #153: [Bug]: bing_webmaster.py uses removed Bing method GetLinkDetails (404) — links/counts/compare all fail
## Summary
`scripts/bing_webmaster.py` calls Bing Webmaster API methods that no longer resolve, so all three Bing commands (`links`, `counts`, `compare`) fail even with a valid API key and a verified site.

## Environment
- claude-seo v2.2.0
- Windows 11, Python 3.14
- Valid Bing Webmaster API key in `~/.config/claude-seo/backlinks-api.json`; target site verified in Bing Webmaster Tools (confirmed via `GetUserSites` -> HTTP 200)

## Steps to reproduce
```
python3 scripts/bing_webmaster.py links https://<verified-site>/ --json
```

## Actual result
```
{
  "status": "error",
  "error": "404 Client Error: Not Found for url: https://ssl.bing.com/webmaster/api.svc/json/GetLinkDetails?siteUrl=...&apikey=***"
}
```
`counts` fails the same way; `get_link_counts()` additionally calls `GetUrlTrafficInfo`, which returns HTTP 400.

## Root cause
`GetLinkDetails` is not a valid Bing Webmaster API method (404). Endpoint probe with a valid key + verified `siteUrl`:

| Method | HTTP |
|---|---|
| `GetUserSites` | 200 |
| `GetLinkCounts` | 200 |
| `GetUrlLinks` | 400 (requires a `link` param) |
| `GetUrlTrafficInfo` | 400 |
| `GetLinkDetails` | **404** |

So `get_link_details()` (used by `links` + `compare`) and `get_link_counts()` (uses `GetUrlTrafficInfo`) both hit non-working endpoints.

## Suggested fix
Repoint `get_link_details()` to `GetLinkCounts` (params `siteUrl` + `page`), which returns `{"d": {"Links": [...], "TotalPages": N}}` — each entry has `Url` + `Count` (no anchor text at this level). Base `get_link_counts()` on `GetLinkCounts` too and drop the `GetUrlTrafficInfo` call. `compare_links()` then works unchanged (it builds on `get_link_details()`).

Verified locally against a real verified property: all three commands return `status: success` after the change.

wonsukchoi:
GetLinkDetails does not exist in the Bing Webmaster API (confirmed against Microsoft's IWebmasterApi reference) and always 404s. GetUrlTrafficInfo 400s the way get_link_counts() called it. All three CLI commands (links, counts, compare) failed even with a valid API key and a verified site. Fixes #153.

## Summary

Verified the root cause against Microsoft's official `IWebmasterApi` reference (learn.microsoft.com): `GetLinkDetails` is absent from the documented method list entirely — there is no such endpoint, which is why it 404s. The two real methods are `GetLinkCounts(siteUrl, page)` (site-wide, page-level inbound link counts — `{Url, Count}` per page, no source/referrer data) and `GetUrlLinks(siteUrl, link, page)` (real referring links with anchor text, but scoped to one specific page URL per call).

The issue's suggested fix — repoint `get_link_details()` straight to `GetLinkCounts` — would stop the 404, but `GetLinkCounts` has no source-URL field. `compare_links()` builds its domain-gap analysis by extracting `source_url` from each link entry; with `GetLinkCounts` alone, `source_url` would always be empty, so `compare_links()` would keep returning `status: success` while silently producing empty domain sets (gap/shared/unique all zero) — a passing-looking result with no real data behind it.

Implemented a two-tier fix instead:
- `get_link_counts()`: sums `Count` across `GetLinkCounts`'s page-0 sample. Drops `GetUrlTrafficInfo` entirely.
- `get_link_details()`: calls `GetLinkCounts` to discover the site's most-linked pages, then expands the top N (default 10, bounds API call count under the existing 1 req/sec rate limit) into real referring links via `GetUrlLinks` per page. `compare_links()` is unchanged but now receives genuine `source_url` data, so its domain-gap math actually works.

Added `tests/test_bing_webmaster.py` (mirrors `test_moz_api.py`'s mocking style) with fixture responses matching Microsoft's documented `GetLinkCounts`/`GetUrlLinks` JSON samples exactly — verifies both endpoints parse correctly, that `GetLinkDetails`/`GetUrlTrafficInfo` are never called, and that `compare_links()` produces a real non-empty gap instead of the hollow-zeros failure mode a naive repoint would hit. Full suite: 330 passed (326 existing + 4 new).

I don't have a live Bing Webmaster API key, so I couldn't reproduce the original reporter's live verification end-to-end; the fix is verified against Microsoft's official API documentation plus mocked unit tests covering the exact documented response shapes.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Tested with a real URL before submitting — no live Bing Webmaster API key available; verified against Microsoft's official API reference and mocked unit tests instead (see Summary).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **bing_webmaster.py links/counts/compare all failed (#153).** `get_link_details()` called `GetLinkDetails`, which does not exist in the Bing Webmaster API (confirmed against Microsoft's `IWebmasterApi` reference) and always 404s; `get_link_counts()` called `GetUrlTrafficInfo`, which 400s the way it was used. Both are now built on `GetLinkCounts` (site-wide, page-level inbound link counts) — `get_link_details()` additionally expands the site's most-linked pages into real referring links via `GetUrlLinks` (Bing has no single endpoint that returns site-wide links with source URLs), so `compare_links()`'s domain-gap analysis keeps working against genuine referrer data instead of silently returning empty sets. Added `tests/test_bing_webmaster.py` with mocked responses matching Microsoft's documented JSON shapes.
+
 ## [2.2.0] - 2026-06-12
 
 Security, cross-platform, and data-accuracy release. Folds the v2.1.0 currency content into the first public ship and closes the full open-issue and PR backlog. No breaking changes.
--- a/scripts/bing_webmaster.py
+++ b/scripts/bing_webmaster.py
@@ -156,52 +156,85 @@ def _normalize_site_url(url: str) -> str:
     return url
 
 
-def get_link_details(site_url: str, api_key: str, page: int = 0) -> dict:
+def get_link_details(site_url: str, api_key: str, page: int = 0, max_pages_to_expand: int = 10) -> dict:
     """
     Get inbound link details for a verified site.
+
+    Bing's Webmaster API has no endpoint that returns site-wide inbound
+    links together with their source URLs (the documented `GetLinkDetails`
+    method does not exist; calling it 404s). This discovers which of the
+    site's pages have inbound links via `GetLinkCounts`, then expands the
+    most-linked pages into individual referring links via `GetUrlLinks`,
+    which requires a specific page URL per call.
 
     Args:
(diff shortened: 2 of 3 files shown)


AgriciDaniel:
Thank you for identifying and working through the removed Bing methods. v2.2.4 ships the supported GetLinkCounts and GetUrlLinks design with bounded pagination, partial failures, sanitized output, and tests. Your initiative directly informed the release.

PR #171, opened 2026-07-18
feat(seo): add /seo setup for post-plugin-install provisioning

### LINKED ISSUE #137: [Bug]: /plugin install does not provision Python deps or Playwright Chromium — fresh plugin install is non-functional until manual setup
## Summary
Installing via the marketplace plugin path (`/plugin marketplace add AgriciDaniel/claude-seo` + `/plugin install claude-seo@agricidaniel-claude-seo`) installs the skills/agents/commands/hooks/MCP config, but it does **not** run the Python provisioning that `install.sh` / `install.ps1` perform — namely `pip install -r requirements.txt` and `python -m playwright install chromium`. On a machine that doesn't already happen to have those deps, every skill that shells out to a `scripts/*.py` (e.g. `render_page.py`, `google_auth.py`, `pagespeed_check.py`) fails at runtime, so a fresh plugin install is effectively non-functional until the user discovers and runs the provisioning manually.

## Environment
- Windows 11, Claude Code
- claude-seo v2.2.0 installed via `/plugin install claude-seo@agricidaniel-claude-seo`
- Python 3.13

## Repro
1. On a clean environment, `/plugin install claude-seo@agricidaniel-claude-seo`.
2. Run any skill that renders/fetches, e.g. `/seo technical <url>`.
3. The underlying `python3 scripts/render_page.py ...` call fails because `playwright` / `trafilatura` / `lxml` etc. are not installed and Chromium was never downloaded.

## Evidence
- The README presents the plugin path as "the fastest path" with no note that Python deps + Chromium must be installed separately; only the manual `install.sh` / `install.ps1` sections run `pip install -r` + `playwright install`.
- After a `/plugin` install, manually running `pip install -r <plugin>/requirements.txt` reported numerous packages as newly installed/upgraded, and `playwright install chromium` had to be run manually before `render_page.py` succeeded.

## Suggested fix
Either:
- (a) Document a required post-install step for the plugin path (`pip install -r <plugin>/requirements.txt` + `python -m playwright install chromium`, including how to locate the plugin's `requirements.txt`), **or**
- (b) Bootstrap it automatically — e.g. a first-run / SessionStart provisioning check, or a `/seo setup` command that installs deps + Chromium when missing.

## Note
Related to, but distinct from, the `python3` interpreter-resolution problem on Windows (filed as a separate issue for cleaner triage).

wonsukchoi:
/plugin install claude-seo@agricidaniel-claude-seo installs skills, agents, commands, and hooks, but unlike install.sh/install.ps1 it never runs pip install -r requirements.txt or playwright install chromium. Every sub-skill that shells out to a script (render_page.py, pagespeed_check.py, capture_screenshot.py, etc.) failed at runtime on a fresh plugin install until the user discovered and ran provisioning manually. Fixes #137.

## Summary

Went with the issue's option (b) — bootstrap it — implemented as a documented `/seo setup` command rather than a silent SessionStart hook. Reasoning: automatically running `pip install`/`playwright install` on every session start without the user asking for it would silently mutate their Python environment, which felt like the wrong default for a hook that fires on every session. A command the user explicitly invokes (matching the repo's existing `/seo backlinks setup` pattern) keeps that as a deliberate action while still solving the "fresh install is silently non-functional" problem, since `seo/SKILL.md`'s Error Handling table now proactively points the model at `/seo setup` the moment a script fails with `ModuleNotFoundError`/`ImportError` or a Playwright missing-browser error — so the user gets the fix suggested right at the point of failure rather than needing to discover it themselves.

Changes:
- `skills/seo/SKILL.md`: new `/seo setup` command in the Quick Reference table, a `## Setup` section describing exactly what it does (locate `requirements.txt` via `${CLAUDE_PLUGIN_ROOT}` for plugin installs or `~/.claude/skills/seo` for manual ones, install deps into the existing venv if `install.sh` created one else `--user`, install Chromium, confirm via import check), and a new Error Handling row.
- `README.md`: the plugin-install section now notes the extra step is required, addressing the issue's evidence point that the README presented `/plugin install` as complete with no such note.

I couldn't test this end-to-end (would require actually running `/plugin install` in a fresh Claude Code environment and invoking `/seo setup`, which isn't available in this session) — this is a documented/model-interpreted command, not code with an automated test surface. Verified the `requirements.txt` packages referenced in the confirmation import check (`playwright`, `trafilatura`, `lxml`) actually match `requirements.txt`, and that the `${CLAUDE_PLUGIN_ROOT}` convention matches what `hooks/hooks.json` already uses for the plugin-install path.

## Type of Change

- [ ] Bug fix
- [x] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Tested with a real URL before submitting — not applicable; this is a documented setup command with no URL/audit surface, and I don't have a fresh /plugin install environment to exercise it end-to-end. See Summary for how it was verified instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **`/seo setup` command (#137).** `/plugin install` installs skills, agents, and hooks but never runs `pip install -r requirements.txt` or `playwright install chromium`, unlike the manual installers — so a fresh plugin install was non-functional until the user discovered and ran provisioning manually. `/seo setup` locates the right `requirements.txt` (`${CLAUDE_PLUGIN_ROOT}` for plugin installs, `~/.claude/skills/seo` for manual ones), installs dependencies and Chromium, and is now proactively suggested whenever a script fails with `ModuleNotFoundError`/`ImportError` or a Playwright missing-browser error. README's plugin-install section now notes the extra step.
+
 ## [2.2.0] - 2026-06-12
 
 Security, cross-platform, and data-accuracy release. Folds the v2.1.0 currency content into the first public ship and closes the full open-issue and PR backlog. No breaking changes.
--- a/README.md
+++ b/README.md
@@ -84,6 +84,11 @@ The fastest path. One-time marketplace add, then plugin install:
 /plugin install claude-seo@agricidaniel-claude-seo
 ```
 
+> **One more step:** unlike the manual installers below, `/plugin install` does not run
+> `pip install -r requirements.txt` or `playwright install chromium` for you. Run
+> `/seo setup` once after installing to provision both — otherwise scripts like
+> `render_page.py` will fail with `ModuleNotFoundError` or a missing-Chromium error.
+
 ### Manual Install (Unix / macOS / Linux)
 
 ```bash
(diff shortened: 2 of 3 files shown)


AgriciDaniel:
Thank you for pushing the setup gap forward. v2.2.4 ships a consolidated setup and doctor runtime with persistent plugin data, rollback-safe upgrades, and cross-platform tests, so this patch is superseded by the released implementation.

PR #179, opened 2026-07-22
test: add coverage for validate_backlink_report

nicokeim1308:
The script validates schema/verification/health-score findings before they reach the user but had no unit tests. Cover the false-positive and false-negative cases it's specifically designed to catch (unflattened @graph schemas, deprecated HowTo, social-media JS false negatives, summary/count mismatches, low-confidence health scores).

## Type of Change

- [ ] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Tested with a real URL before submitting
- [ ] SKILL.md files stay under 500 lines (if modified)
- [ ] Python scripts output JSON (if modified)
- [ ] Reference files stay under 200 lines (if modified)
- [ ] `set -euo pipefail` used in any new shell scripts
- [ ] CHANGELOG.md updated with the change

PR #178, opened 2026-07-21
fix(seo-technical): dynamic rendering is a workaround, not a target state

hwcgav:
## Summary

The JavaScript Rendering section of `seo-technical` lists **"Verify dynamic rendering setup if applicable"** as an audit step. That reads as an endorsement of dynamic rendering, but Google documents it as a workaround rather than a recommended approach:

> Dynamic rendering is a workaround and not a recommended solution, because it creates additional complexities and resource requirements.

— [Search Central: Dynamic rendering as a workaround](https://developers.google.com/search/docs/crawling-indexing/javascript/dynamic-rendering) (page last updated 2025-12-10)

The same page notes dynamic rendering "**was** a workaround" and points to server-side rendering, static rendering, or hydration as better options.

This changes the step to flag detected dynamic rendering as technical debt and recommend those alternatives, with a link to the source.

**Detection logic is deliberately untouched.** `scripts/fetch_page.py` compares Googlebot-UA responses against the default UA to spot prerender services — that remains a valid SPA signal and is unaffected. Only the recommendation was stale.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [x] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [x] Tested with a real URL before submitting — verified against the Search Central source page; no code paths changed, so no behavioural test applies
- [x] SKILL.md files stay under 500 lines (228 after the change)
- [x] Python scripts output JSON (if modified) — no scripts modified
- [x] Reference files stay under 200 lines (if modified) — no reference files modified
- [x] `set -euo pipefail` used in any new shell scripts — no shell scripts added
- [x] CHANGELOG.md updated with the change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #182, opened 2026-07-26
fix(tests): stop module-level sys.exit from aborting the whole pytest session

Nordalux:
## Summary

`scripts/gsc_query.py` calls `sys.exit(1)` at module scope when `google-api-python-client` is missing. Three test modules import it at module scope, so pytest raises `SystemExit` during collection, which becomes an `INTERNALERROR` and aborts the whole session — not three modules, all of them. `tests/test_sitemap_discovery.py` does the same via `lxml`.

Adds a `pytest.importorskip()` guard to the four affected modules. The scripts are untouched, so their CLI error messages stay as they are.

```bash
pip install pytest beautifulsoup4
pytest tests/ -q
```

Before: `INTERNALERROR ... SystemExit: 1`, `no tests ran`, exit 3.
After: `371 passed`.

Verified in:
- Windows 11, Python 3.14, clean venv, `pytest` + `beautifulsoup4` only — 371 passed, 4 skipped (15 pre-existing failures are POSIX permission assertions that cannot hold on NTFS, untouched here)
- Same checkout with the full `requirements.txt` — the four modules execute, 24 passed, so nothing is silently skipped where the dependency exists
- Your CI on `main` @`09d37c7` — green, confirming those 15 pass on Linux

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/tests/test_gsc_pagination.py
+++ b/tests/test_gsc_pagination.py
@@ -6,8 +6,16 @@ import sys
 from pathlib import Path
 from unittest import mock
 
+import pytest
+
 ROOT = Path(__file__).resolve().parents[1]
 sys.path.insert(0, str(ROOT / "scripts"))
+
+# gsc_query calls sys.exit(1) at import time when google-api-python-client is
+# missing. Under pytest that SystemExit escapes collection as an INTERNALERROR
+# and aborts the whole session, so no test in tests/ runs. Skip this module
+# instead when the optional Google stack is absent.
+pytest.importorskip("googleapiclient")
 
 import gsc_query  # noqa: E402
 
--- a/tests/test_gsc_query.py
+++ b/tests/test_gsc_query.py
@@ -7,11 +7,19 @@ import sys
 from pathlib import Path
 from unittest.mock import patch
 
+import pytest
+
 
 REPO_ROOT = Path(__file__).resolve().parents[1]
 _SCRIPTS = REPO_ROOT / "scripts"
 if str(_SCRIPTS) not in sys.path:
     sys.path.insert(0, str(_SCRIPTS))
+
+# gsc_query calls sys.exit(1) at import time when google-api-python-client is
+# missing. Under pytest that SystemExit escapes collection as an INTERNALERROR
+# and aborts the whole session, so no test in tests/ runs. Skip this module
+# instead when the optional Google stack is absent.
+pytest.importorskip("googleapiclient")
 
 import gsc_query  # noqa: E402
 
--- a/tests/test_gsc_totals_aggregate.py
+++ b/tests/test_gsc_totals_aggregate.py
@@ -9,8 +9,16 @@ import sys
 from pathlib import Path
 from unittest import mock
 
+import pytest
+
 ROOT = Path(__file__).resolve().parents[1]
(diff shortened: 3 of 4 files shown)


AgriciDaniel:
This one deserves to go in first — it's more consequential than the diff size suggests.

While reviewing the other open PRs I ran the suite on each branch. On **every** branch except this one, pytest dies during collection and **no tests run at all**:

```
INTERNALERROR> SystemExit: 1
mainloop: caught unexpected SystemExit!
no tests ran in 0.44s
```

`scripts/gsc_query.py` calls `sys.exit(1)` at import time when `google-api-python-client` is missing, and that `SystemExit` escapes collection and kills the whole session — not one module, all of them. CI hides it because CI installs every optional dependency.

On this branch: `232 passed`. (The one failure I saw, `test_sync_flow.py::test_dry_run_exits_zero`, makes a live `urlopen` call to the GitHub API — environmental, unrelated to this PR.)

So until this lands, a green local run proves nothing and contributors can't validate their own work. Thanks for finding it.

---

AgriciDaniel:
Recording the suggested merge order here, since this PR is the one everything else depends on. Full write-up with evidence is in `docs/REVIEW-open-prs-issues-2026-08.md` on `claude/prs-issues-review-1zwegz`.

**1. This PR (#182)** — nothing else is verifiable until it lands. Without it pytest aborts at collection and zero tests run in any environment missing the optional Google stack.

**2. #183** — one line, removes `|| true` from the v2 dependency install. Same failure mode from the other direction: a swallowed install currently resurfaces as an opaque INTERNALERROR rather than the pip error that caused it.

**3. #241** — worth prioritising higher than its size suggests. Yoast emits a top-level `@graph` on every page of every site running it, so the `Missing @type` false positive rejects the standard output of the most widely deployed SEO plugin on the web, and blocks the user's edit. I probed it adversarially: missing `@context`, a member without `@type`, and a bad `@context` all still error correctly, so it doesn't over-permit.

**4. #240** — hold briefly. The bug it fixes is real, but the new pattern blocks ordinary product copy ("Replace the cartridge every 6 months") while letting `REPLACE_ME` and `REPLACE_WITH_URL` through. Details and a suggested one-line pattern are on that PR.

**Then:** Dependabot #190–#194 (#193's numpy 2.x floor is safe here — the only numpy APIs used anywhere are `np.linspace` and `np.pi`), close #201 as superseded by #190, and #185 / #178 / #179 / #195 whenever convenient.

**Not ready:** #198 (leaves `scripts/bing_webmaster.py` unparseable), #197/#203 (unvalidated fetch of a URL from an API response), #204 (right destination, but can corrupt `~/.claude.json`), #209 (superseded by `claude-seo run`).

Thanks to everyone who filed these — several turned out to be more consequential than they looked.

---

AgriciDaniel:
Landed in v2.2.6 as commit db4b68 (cherry-picked with -x, your authorship preserved). Thank you. Landed as-is; the 13 remaining scripts with import-time sys.exit are tracked for v2.3.0.

PR #183, opened 2026-07-26
fix(ci): do not swallow requirements install failure in v2 audit

Nordalux:
## Summary

`v2.yml`, job `full_pytest`, installs the runtime with `pip install -r requirements.txt || true`. A failed install stays invisible; the run then dies later as `INTERNALERROR ... SystemExit: 1` / `no tests ran`, because several test modules import scripts that exit at module scope when a dependency is absent. The log never mentions pip.

Drops the `|| true`, matching `ci.yml`, which installs the same set without a fallback.

Verified in:
- YAML parses clean (`yaml.safe_load`) on Windows 11, Python 3.14
- Not exercisable from a PR: the workflow is `workflow_dispatch` / `workflow_call` only, so it wants a manual run from the Actions tab

Two adjacent things I left alone: the header comment still says *"Until the v2 branch is pushed"*, and `secret_scan` is byte-identical to `secret-scan` in `ci.yml`. Happy to open follow-ups.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/.github/workflows/v2.yml
+++ b/.github/workflows/v2.yml
@@ -84,9 +84,14 @@ jobs:
           python-version: "3.10"
 
       - name: Install pytest + test deps
+        # Must not be tolerated with `|| true`: several test modules import
+        # scripts that call sys.exit(1) at module scope when a dependency is
+        # missing, so a swallowed install failure resurfaces later as an opaque
+        # pytest INTERNALERROR instead of the pip error that caused it. Matches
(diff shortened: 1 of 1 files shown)


AgriciDaniel:
Landed in v2.2.6 as commit 23f94b (cherry-picked with -x, your authorship preserved). Thank you.

PR #195, opened 2026-07-27
Add templated-metadata detector: catch generated title/description pairs

sulemank137:
## Summary

Adds a **templated-metadata detector** — `scripts/metadata_template.py` — that flags
machine-generated `<title>` / `<meta name="description">` pairs.

The signature is narrow: a description that opens by restating its own title verbatim
and then closes with a stock call to action ("Try it free now.", "Start free!",
"Learn more now!"). Bulk metadata jobs — a CSV column piped through a template, or an
LLM asked for "a description for each of these titles" — emit that exact shape on
every URL of a site at once. That site-wide repetition is what turns it from a
copywriting nit into a scaled-content signal.

Detection is four deterministic string comparisons. No model, no inference about
authorship, nothing that could be described as an "AI detector".

| Signal | Severity | Fires when |
|---|---|---|
| `templated_metadata` | high | Description echoes the title, then ends on a stock CTA |
| `description_echoes_title` | medium | Description echoes the title but adds real information |
| `brand_suffix_in_description` | low | The title's `\| Brand` suffix was concatenated into the description body |
| `description_duplicates_title` | medium | Description and title are the same string |

`analyse_pairs()` rolls single pages up to a site verdict (`templated_ratio`,
`shared_cta_phrases`, `site_risk`), because the spam signal operates at site scale:
one templated description is an oversight, the same closing phrase on a third of the
site is a pattern.

## Why this is a new check and not a note on an existing one

I searched the repo before writing anything, because the last thing a maintained
project needs is a duplicate check. **Nothing in claude-seo currently compares a meta
description against its own title.** What exists and why each one misses this:

- `scripts/content_quality.py` — scores body text (filler, AI phrasings, information
  density, repetition). Never sees metadata.
- `scripts/drift_compare.py` — rule 9 compares the title against *its own past value*,
  rule 10 does the same for the description. Neither compares the two to each other.
- `skills/seo-programmatic/SKILL.md` — the uniqueness gate is
  `words unique to this page / total words on page`, explicitly body copy. A set of
  pages with fully unique body text passes at 100%.
- `skills/seo-page/SKILL.md` and `skills/seo/references/quality-gates.md` — check the
  description for length, keyword presence, and cross-page uniqueness. "Unique per
  page" is necessary but not sufficient: 26 descriptions can each be a unique string
  while all sharing one generated shape.
- `skills/seo-content/SKILL.md` lists "repetitive structure across pages" as a
  low-quality AI marker, but as a judgement call for the model, with nothing
  deterministic behind it and no metadata scope.

Greps run, all empty: `duplicat.*title`, `restate`, `echo the title`,
`matches the title`, `same as title`, `identical to the title`, `title duplication`,
plus a scan of every script that reads both `title` and `meta_description`
(`parse_html.py` parses both without comparing; `drift_baseline.py` stores both;
`drift_compare.py` diffs each against history).

## The incident this comes from

This is not a hypothetical. A live site (credify.work) was demoted by the **June 2026
Google spam update** for scaled content abuse. The cause was this exact metadata
pattern across 26 pages, produced by a bulk CSV metadata job.

The important part: **the body copy was genuinely unique** — average cross-page 8-gram
Jaccard similarity of 0.021. Every body-content heuristic passed. A full claude-seo
audit was run against the site and did not surface the problem, because every check
that could have caught it was looking at body copy.

That is the gap this PR closes, and it is why the detector is also wired into the
`seo-programmatic` uniqueness gate — that gate measures body words by design, which is
precisely the route by which this slips through.

## One upstream bug found and fixed in the port

The reference implementation I ported from had a regex gap: `"Start your free scan
now."` never matched its CTA branch. It went unnoticed because both the CTA branch and
the no-CTA branch emitted the same signal id, so the QA suite's `"templated_metadata"
in signals` assertion passed via the weaker branch. The port generalises the CTA
patterns (bounded-gap forms like `start (your|the)? \w{0,3} (free|now)`) and, more
importantly, the tests here assert on **severity**, not just presence, so the same
class of bug cannot hide again.

## Wiring

- `skills/seo-page/SKILL.md` — single-URL on-page checks
- `agents/seo-content.md` — the audit path, with both single-page and `--pairs-file` usage
- `skills/seo/references/quality-gates.md` — new "not a title restatement" row, plus a
  note qualifying the existing "include compelling CTA" requirement (a CTA written for
  *this* page, not one appended to every page by a template)
- `skills/seo-programmatic/SKILL.md` — metadata is scored separately from body uniqueness
- `CLAUDE.md` / `AGENTS.md` script list and count (53 → 54)
- `CHANGELOG.md` under `[Unreleased]`

## Tests

`tests/test_metadata_template.py`, written in the existing `test_content_quality.py`
style (pytest, `parametrize`, `sys.path` insert of `scripts/`).

- **6 true positives** — the actual metadata from the demoted pages
- **8 false positives** — this is the half that matters, and it breaks down as:
  - **5 pages that legitimately open with their own topic phrase**, because that is
    simply how a sentence about that topic starts ("Word counter that gives instant
    word, character and sentence counts…"), plus one where the brand name is genuinely
    the subject of the page. All come back completely clean.
  - **1 page that ends on a real CTA without echoing its title.** A CTA on its own is
    never evidence; the test asserts the CTA regex *does* match the description and that
    the page is still not flagged.
  - **2 pages that echo their title and then end on "now" / "free" in the ordinary
    sense** ("…what a passing score actually means now.", "…check 200 URLs for free.").
    These guard the broadened CTA patterns. They correctly receive the medium
    `description_echoes_title` advisory and are correctly **not** flagged as templated.
- **10 edge cases** — empty strings, `None`, single characters, a bare separator, a
  500-character title, HTML/script content, accented and em-dash titles
- **CLI contract** — JSON validity, `--fail-on` exit codes, `--pairs-file` including
  `parse_html.py`'s `meta_description` key
- **Site roll-up** — high/medium/isolated thresholds, shared-CTA counting, empty input

Titles below a normalised 25-character floor are exempt from the echo test, which is
what keeps the "Word Counter" class of page clean.

### Results

```
Baseline (main, before this PR):  410 passed
With this PR:                     458 passed  (+48 new)
```

No existing test was modified or removed. Also verified against the repo's own gates:

```
scripts/consistency_check.py     PASS: 0 errors, 0 warnings, 381 files
scripts/portability_check.py     33 SKILL.md files, 0 errors, 0 warnings
CI py_compile (scripts/*.py)     All 54 scripts passed syntax check
ruff check scripts/metadata_template.py   clean
```

(`ruff check tests/` reports `I001` on the new test file, as it already does on
`test_content_quality.py` and eight other existing test files — the
`sys.path.insert`-then-import idiom inherently triggers it. I matched the house style
rather than diverge from it. Happy to add a `per-file-ignores` entry instead if you'd
prefer.)

## Type of Change

- [x] New feature / sub-skill

## Checklist

- [x] Tested with a real URL before submitting — the detector takes title/description
      strings rather than fetching, so it was exercised against the real metadata of
      the 26 demoted pages plus the CLI end-to-end
- [x] SKILL.md files stay under 500 lines (seo-page 100, seo-programmatic 180)
- [x] Python scripts output JSON (`--json`, plus a human-readable default)
- [x] Reference files stay under 200 lines (quality-gates.md 166)
- [x] No new shell scripts

## Notes for review

- Thresholds (`_MIN_TITLE_CHARS`, `_SITE_HIGH_RATIO`, and the rest) are named module
  constants with the reasoning in comments, so they're easy to retune.
- `_CTA_TAIL_PATTERNS` is deliberately conservative. The docstring asks that additions
  be justified by a phrase seen in real generated metadata rather than intuition,
  mirroring the note already on `_AI_PATTERNS` in `content_quality.py`.
- The detector never claims content is AI-written. It reports a structural property of
  two strings, which keeps it defensible in a client-facing report.
- **The `quality-gates.md` edit is the one change I'd expect pushback on.** It qualifies
  your existing "Include compelling CTA" row to read "written for this page", because a
  templated CTA is part of the pattern this detects and the unqualified guidance can be
  read as encouraging it. That is your published advice, not mine — if you'd rather keep
  the row as it was and let the new "not a title restatement" row carry the point on its
  own, say so and I'll drop that hunk. The detector works either way.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -56,7 +56,7 @@ in case a recipe needs a specific call.
 
 Claude SEO is a Tier 4 SEO analysis skill with 25 sub-skills (21 core + 1 orchestrator +
 1 framework integration + 2 extension mirrors), 18 sub-agents (15 core + 1 framework
-integration + 2 extension mirrors), and 53 Python execution scripts.
+integration + 2 extension mirrors), and 54 Python execution scripts.
 
 ## Quick Reference
 
@@ -165,7 +165,7 @@ skills/                    # 25 sub-skills (auto-discovered)
   seo-dataforseo/         # DataForSEO (extension)
   seo-image-gen/          # AI images (extension)
 agents/                    # 18 subagents
-scripts/                   # 53 Python scripts, including the managed runtime
+scripts/                   # 54 Python scripts, including the managed runtime
 schema/                    # JSON-LD templates
 extensions/                # 8 MCP extensions: DataForSEO, Firecrawl, Banana, Ahrefs, SE Ranking, Profound, Bing Webmaster, Unlighthouse
 ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Templated-metadata detector (`scripts/metadata_template.py`): flags meta descriptions that
+  restate their own title tag verbatim and then close with a stock call to action, the shape bulk
+  metadata jobs produce site-wide. Deterministic string comparison, no model. Exposes a site-level
+  roll-up (`templated_ratio`, `shared_cta_phrases`, `site_risk`) because scaled content abuse is a
+  site-scale signal, not a per-page one. Wired into `seo-page`, the `seo-content` agent, the
+  quality-gates meta description table, and the `seo-programmatic` uniqueness gate, which measures
+  body copy only and therefore cannot see this. 48 new tests.
+
 ## [2.2.4] - 2026-07-20
 
 Community maintenance release following a full review of every open issue and pull request.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -85,7 +85,7 @@ claude-seo/
     seo-flow.md                  # FLOW framework integration
   hooks/                           # Quality gate hooks
     hooks.json                   # PostToolUse schema validation
-  scripts/                         # 53 Python execution scripts
+  scripts/                         # 54 Python execution scripts
     google_auth.py               # Credential management (OAuth, SA, API key, 4-tier detection)
     backlinks_auth.py            # Backlink API credential management (Moz, Bing)
     moz_api.py                   # Moz Link Explorer API (DA/PA, spam, domains, anchors)
@@ -120,6 +120,7 @@ claude-seo/
     preload_check.py             # Speculation Rules / bfcache / prerender / preload detector
     agent_ux_check.py            # Agent-friendly page auditor
     content_quality.py           # QRG-aligned content quality detector
+    metadata_template.py         # Templated title/description detector (title echo + stock CTA)
     content_humanize.py          # AI-pattern remover (rewrites AI-typical phrasing)
     content_verify.py            # Claim extractor + citation-gap detector
     schema_generate.py           # JSON-LD generators for high-leverage v2 schema types
--- a/agents/seo-content.md
+++ b/agents/seo-content.md
@@ -17,6 +17,32 @@ When given content to analyze:
 5. Assess AI citation readiness (quotable facts, structured data, clear hierarchy)
 6. Check content freshness and update signals
 7. Flag potential AI-generated content quality issues per Sept 2025 QRG criteria
+8. Check title/description pairs for templating (see below)
+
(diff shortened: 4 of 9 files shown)


AgriciDaniel:
Useful heuristic with strong tests. Two changes before it ships: label the output as a heuristic (the word appears nowhere in the script or its JSON; content_quality.py and agent_ux_check.py both label theirs), and remove the claim that templated metadata caused the August spam update, which is unverified and appears in the module docstring, the high-severity message, seo-page, seo-programmatic, and quality-gates.md, not only in the PR body.

AgriciDaniel:
Landed in v2.3.0 as commit 270e247; output is labelled method: heuristic and the spam-update causality claim was removed. Thank you.

PR #196, opened 2026-07-29
chore: ponytail cleanup — 5 findings (-102 lines)

### LINKED ISSUE #4: Opencode install please :)

pookNast:
## Summary

Ponytail-over-engineering cleanup pass — removes 102 lines of dead code, unused imports, and duplicated logic. Found by running the [`/ponytail-review`](https://github.com/DietrichGebert/ponytail) audit on a vendored copy of this skill.

All 5 findings are mechanical cleanups; behavior is unchanged. No public-API surface removed that had a real caller.

### 1. `merge:` duplicated `_sha256` in `release_sign.py:91` + `verify_release.py:42`

Both scripts had identical 5-line chunk-loaded SHA-256 implementations. Collapsed to 2 lines each using the Python 3.11+ stdlib `hashlib.file_digest` helper (which does chunked loading internally).

```python
def _sha256(path: Path) -> str:
    with path.open("rb") as fh:
        return hashlib.file_digest(fh, "sha256").hexdigest()
```

**Note:** requires Python 3.11+. If you support older Python, happy to keep the duplication and drop this finding from the PR.

### 2. `shrink:` `urlparse` re-imported 3× inside function bodies (`validate_backlink_report.py:104,184,195`)

Hoisted to a single module-top import. Net -3 lines.

### 3. `delete:` uncalled placeholder normalizers in `dataforseo_normalize.py:310-374`

`normalize_social` and `normalize_reviews` are documented placeholders ("for future seo-social skill integration", "for future expansion"). Neither has a caller anywhere in the skill. They were wired into the CLI `--module` choices, so removing them also drops the `social` and `reviews` choices from `--module`.

**Behavior change:** `--module social` and `--module reviews` now fail with an argparse error instead of returning `[]`. Since these were placeholders for skills that don't exist, no real user should be calling them — but flagging in case you'd rather keep them as no-ops.

### 4. `delete:` dead `_stream_gz_lines` in `commoncrawl_graph.py:130-145`

Zero callers (grep confirmed). The docstring claims "stream and decompress on the fly to avoid loading multi-GiB files into memory" but `io.BytesIO(resp.content)` materializes the entire response before `GzipFile` can stream — the docstring lies. The sibling `_stream_gz_chunked` below uses `import zlib` + incremental decompression and is the actual streaming implementation.

### 5. (Originally part of the audit, already fixed on `main` separately)

The audit also flagged an unused `quote` import in `bing_webmaster.py`, but `main` already has just `from urllib.parse import urlparse` (no `quote`). So nothing to do here — noting for completeness.

## Test plan

- [x] `python3 -m py_compile` passes for all 5 modified scripts
- [x] `grep -rn 'normalize_social\|normalize_reviews\|_stream_gz_lines' scripts/` returns zero matches (no orphaned callers)
- [x] `grep -rn 'normalizers\[args' scripts/` returns zero matches (orphaned dict lookup removed)
- [ ] Maintainer smoke-test: run any workflow that exercises `release_sign.py` / `verify_release.py` to confirm `hashlib.file_digest` output matches the old chunk-loaded implementation byte-for-byte (it should — same SHA-256 over the same bytes)
- [ ] Maintainer smoke-test: confirm no real users were passing `--module social` or `--module reviews` to `dataforseo_normalize.py`

## Why

Smaller surface = less to maintain, less to read, fewer places for bugs to hide. None of these findings are urgent — just routine cleanup. Happy to split into separate PRs (e.g. bug-fix #4 separate from refactor #1-#3) if that's easier to review.

### DIFF
--- a/scripts/commoncrawl_graph.py
+++ b/scripts/commoncrawl_graph.py
@@ -127,24 +127,6 @@ def _save_cache(domain: str, release: str, data: dict) -> None:
         json.dump(data, f, indent=2)
 
 
-def _stream_gz_lines(url: str, timeout: int = 120):
-    """
-    Stream and decompress a gzipped text file line by line.
-
-    Downloads the file in chunks and decompresses on the fly to avoid
-    loading multi-GiB files into memory.
-
-    Yields:
-        Decoded text lines.
-    """
-    resp = requests.get(url, stream=True, timeout=timeout)
-    resp.raise_for_status()
-
-    decompressor = gzip.GzipFile(fileobj=io.BytesIO(resp.content))
-    for line in io.TextIOWrapper(decompressor, encoding="utf-8"):
-        yield line.rstrip("\n")
-
-
 def _stream_gz_chunked(url: str, target_domain: str, timeout: int = 120,
                         max_lines: int = 0) -> list:
     """
--- a/scripts/dataforseo_normalize.py
+++ b/scripts/dataforseo_normalize.py
@@ -304,77 +304,6 @@ def normalize_merchant(items: list[dict[str, Any]]) -> list[dict[str, Any]]:
 
 
 # ---------------------------------------------------------------------------
-# Social normalizer (placeholder for future seo-social skill)
-# ---------------------------------------------------------------------------
-
-def normalize_social(items: list[dict[str, Any]]) -> list[dict[str, Any]]:
-    """
-    Normalize social signal data from DataForSEO.
-
-    Placeholder for future seo-social skill integration. Handles social
-    media engagement metrics, share counts, and platform-specific data.
-
-    Args:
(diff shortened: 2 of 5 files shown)


AgriciDaniel:
Two of your hunks (import hoisting, dead-code removal) are wanted. hashlib.file_digest is Python 3.11+ while pyproject declares >=3.10, and removing the --module choices breaks the documented CLI. If you split the safe hunks into their own PR they can land in v2.3.0.

AgriciDaniel:
AgriciDaniel/claude-seo

pookNast:
Thanks for the review — agreed on both counts (file_digest is 3.11+ vs pyproject >=3.10, and the --module choices are documented CLI surface). Saw 67c2838 applied the safe hunks, so this is fully resolved. Will keep the two rejected findings out of future passes.

PR #200, opened 2026-07-31
fix(url_safety): send browser-like default headers on safe fetches

EPIXSMedia:
## What's wrong

`safe_requests_get` and `safe_requests_head` in `scripts/url_safety.py` never pass a `headers=` mapping, so every raw fetch goes out as:

```
User-Agent: python-requests/2.34.2
```

with no `Accept-Language`. Managed WAFs and CDNs answer that with 403/406, and SSR frameworks answer it with the empty client-side shell.

The failure is silent. Nothing raises — the caller gets a 200-shaped error page or a hydration shell and analyses it as if it were the real document, so an incorrect score comes back looking like a correct one.

Reproducer on current `main`:

```python
from url_safety import safe_requests_get
r = safe_requests_get("https://en.wikipedia.org/wiki/Search_engine_optimization", timeout=30)
print(r.status_code, r.request.headers["User-Agent"])
# 403 python-requests/2.34.2
```

Roughly 11 call sites across 8 scripts are affected, including `render_page.py`, `parse_html.py`, `preload_check.py`, `ucp_check.py` and `parasite_risk.py`.

This is the same class of bug as #9, which was fixed in `fetch_page.py` in v1.2.1. The v2.0.0 `url_safety.py` helpers are a newer code path that never picked up those defaults.

## The fix

Adds `DEFAULT_REQUEST_HEADERS` (the same browser-like values `fetch_page.py` already uses) merged in via `_with_default_headers()`, so **caller-supplied headers still win**.

Deliberately unaffected:

- `fetch_page.py --user-agent` — Googlebot cloaking checks still send Googlebot
- `sitemap_discovery.py` — keeps its `application/xml` Accept header
- `safe_requests_session` — both callers already set their own headers, left alone

URL validation, DNS pinning and SSRF behaviour are untouched; the change only fills in headers.

## Verification

Full suite: **414 passed** (410 existing + 4 new).

One existing test, `test_safe_requests_head_uses_strict_validation_and_dns_pin`, asserted the exact call args and so encoded the missing-headers behaviour. Updated to expect the defaults; its actual purpose (strict validation + DNS pin) is unchanged.

Live before/after:

| URL | before | after |
|---|---|---|
| en.wikipedia.org/wiki/Search_engine_optimization | 403 | **200** |
| a Hostinger/cPanel ModSecurity host | 406 | **200** |
| nytimes.com | 200 | 200 |
| stackoverflow.com | 200 | 200 |

Sites that already worked are unaffected.

Closes the `url_safety.py` half of #9.

### DIFF
--- a/scripts/url_safety.py
+++ b/scripts/url_safety.py
@@ -424,6 +424,34 @@ def _pin_dns(hostname: str, pinned_ip: str, port: int) -> Iterator[None]:
         _dns_patch_lock.release()
 
 
+# Default request headers, mirroring the browser-like defaults fetch_page.py
+# has used since v1.2.1 (issue #9). Without them ``requests`` announces itself
+# as ``User-Agent: python-requests/x.y.z`` with no Accept-Language, which many
+# managed WAFs and CDNs answer with 403/406, and which SSR frameworks answer
+# with the empty client-side shell. Callers do not see an exception in that
+# case, they analyse the error page or the shell as if it were the real
+# document. Any header here can be overridden by passing ``headers=``.
+DEFAULT_REQUEST_HEADERS = {
+    "User-Agent": (
+        "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 "
+        "(KHTML, like Gecko) Chrome/150.0.7871.115 Safari/537.36 ClaudeSEO/2.0"
+    ),
+    "Accept": (
+        "text/html,application/xhtml+xml,application/xml;q=0.9,"
+        "image/avif,image/webp,*/*;q=0.8"
+    ),
+    "Accept-Language": "en-US,en;q=0.9",
+}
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
No SSRF concern, and the intent is sound. Two corrections: the body says these are the same values fetch_page.py already uses, but platform, Chrome patch digit, Accept, and Accept-Language q values differ; and forcing Accept-Language: en-US on 8 previously header-less call sites is untested against the hreflang skill, which depends on the served language variant. Please align the values with fetch_page.py, leave Accept-Language unset unless the caller passes one, and add a test that a request through safe_requests_get carries no Accept-Language by default.

AgriciDaniel:
Landed in v2.3.0 as commit 484ce21; header values are shared with fetch_page.py and Accept-Language is not sent unless the caller passes one. Thank you.

PR #202, opened 2026-08-04
fix: cost ledgers lose concurrent writes (DataForSEO + Banana)

qant-au:
## Summary

Both of the plugin's cost ledgers silently drop entries under concurrent writes and
under-report spend. Same defect, two files, one fix applied consistently to each.

### The bug

`scripts/dataforseo_costs.py` takes its file lock twice, once to read and once to write,
and releases it in between:

```python
def cmd_log(args):
    ledger = _load_ledger()      # LOCK_SH, read, unlock
    ...
    ledger["entries"].append(entry)
    _save_ledger(ledger)         # LOCK_EX, write, unlock
```

Two concurrent `log` calls both read the same ledger, and the second overwrites the first:

```
P1 load -> [a,b,c]
P2 load -> [a,b,c]
P1 append d, save -> [a,b,c,d]
P2 append e, save -> [a,b,c,e]     # d is gone
```

Upgrading `_load_ledger` to `LOCK_EX` would not fix this. The lock has to span both halves.
`cmd_reset` has the same shape, so a concurrent `log` during a reset is lost, or the reset is.

**Measured on macOS**, 20 concurrent `log` calls against an isolated ledger: 3 entries lost,
and 1 to 3 lost on every repeat run. Originally reported on Windows, which is where it was
first noticed, but it is not platform-specific.

This matters more than the raw numbers suggest. `skills/seo-audit` fans out to up to 15
parallel specialists, three of which call DataForSEO (`seo-dataforseo`, `seo-backlinks`,
`seo-ecommerce`), and `skills/seo-dataforseo/SKILL.md` instructs each to `log` after every
call. Concurrent logging is the designed operating mode, not an edge case, so the ledger
under-reports precisely during the runs that spend the most.

### Two related defects on the same surface

**No file locking at all on Windows.** The module sets `fcntl = None` when the import fails
and then takes every unlocked code path, so on Windows the ledger runs with no locking
whatsoever, and a reader can observe a half-written file.

**A corrupt ledger silently resets the budget to zero.** `_save_ledger` opens the ledger with
`"w"`, which truncates in place and is not atomic. If the process dies mid-write, the file is
left as invalid JSON. `_load_ledger` then hits
`except (json.JSONDecodeError, IOError): return {"entries": []}` and reports an empty ledger,
which the next `log` persists. The entire spend history disappears, not just one entry. That
is strictly worse than the race and shares the same fix surface.

### The fix

- **`_locked_ledger()`** holds one exclusive lock across the whole read-modify-write.
  `cmd_log` and `cmd_reset` both use it. The lock stays in a separate `.lock` file on
  purpose: the atomic write replaces the ledger's inode, so a lock held on the ledger itself
  would not protect the replacement.
- **Windows no longer runs unlocked.** `fcntl` falls back to `msvcrt.locking`, and when
  neither primitive is available the operation raises rather than proceeding unlocked. A
  budget ledger that silently drops entries is worse than one that refuses to run.
- **Ledger writes are atomic** (tempfile plus `os.replace`, the idiom already used in
  `scripts/google_auth.py`), and an unparseable ledger now fails closed with a clear message
  and the file left in place for inspection, instead of restarting the budget at zero.
- **`CLAUDE_SEO_CONFIG_DIR`** overrides the config and ledger location so the tests run
  against an isolated ledger and never touch the operator's real spend history.

### Tests

New `tests/test_dataforseo_costs.py`, 7 tests:

- 20 concurrent `log` subprocesses must produce exactly 20 entries with the correct total.
  This fails on the current code and passes with the fix.
- Interleaved `log` and `reset` must leave the ledger parseable.
- A truncated ledger must fail closed rather than being silently zeroed.
- The atomic write must not leave temp files behind.
- A portability guard asserting locking is never disabled unconditionally, so the
  `fcntl = None` regression cannot come back.

## Part 2: `extensions/banana/scripts/cost_tracker.py`

The image cost ledger has the same defect and is worse: it had **no file locking at all**
on any platform, plus the same non-atomic write.

Measured against v2.2.4, 20 concurrent `log` calls: **4 entries lost**, and **3 of those
processes crashed outright** with `JSONDecodeError` after reading a half-written file. So
on the current code a concurrent batch can both under-report spend and kill generations
mid-run.

It is also harder to notice than the DataForSEO case. This ledger carries running
aggregates (`total_cost`, `total_images`, `daily`) that are incremented in the same write
as the entry, so a lost write drops the entry **and** its aggregate increments together.
The file therefore stays internally self-consistent while under-reporting. Checking that
`total_cost` equals `sum(entries)` proves nothing.

Same treatment as Part 1: one exclusive lock across the whole read-modify-write in
`cmd_log`, atomic writes, `msvcrt` fallback with a hard failure when no locking primitive
exists, and a corrupt ledger that fails closed with a clear message instead of a bare
traceback.

One deliberate asymmetry: `cmd_reset` takes the exclusive lock but does **not** read the
ledger first. Reset is the documented recovery path for a corrupt ledger, so making it
depend on the existing file parsing would leave a corrupt ledger unrecoverable through the
CLI. There is a test for exactly that.

`BANANA_HOME` overrides the ledger and pricing paths so the 9 new tests in
`tests/test_banana_cost_tracker.py` never touch the operator's real `~/.banana`. They
cover concurrency, aggregate correctness after concurrent writes, interleaved
`log`/`reset`, corrupt-ledger fail-closed, reset-recovers-corrupt, no leftover temp files,
tolerance of older ledgers missing the aggregate keys, and a portability guard.

## Verification

Full suite: **412 passed** on this branch, re-run against `main` @ `09d37c7`. The 3 excluded
modules (`test_gsc_pagination.py`, `test_gsc_query.py`, `test_gsc_totals_aggregate.py`) are
gated on `google-api-python-client`, which is not installed in this environment and is
unrelated to this change.

Note for reviewers: the Banana half was written against this repo's current
`extensions/banana/scripts/cost_tracker.py`, which loads the dated `pricing.json`. It is not
a port of an older shape.

## Not addressed here

`check` (read spend) -> API call -> `log` (write spend) is a separate TOCTOU across
processes: N parallel agents can each pass the daily-limit check before any of them logs, so
the daily limit can still be overshot even with a correct ledger. Closing that needs a
reserve/commit design and felt out of scope for a bug fix. Happy to follow up if you want it.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Tested with a real URL before submitting — **not applicable**, and left unticked
  rather than claimed. This change touches only the two spend ledgers; no analysis or
  fetch path is involved, so there is no URL that would exercise it. The equivalent
  evidence is the full suite (412 passed) plus the two new concurrency suites, each of
  which reproduces its race against the current code and shows it absent after the fix.
- [x] SKILL.md files stay under 500 lines (if modified) — none modified.
- [x] Python scripts output JSON (if modified) — preserved. `dataforseo_costs.py` still
  emits JSON on every path, and the new fail-closed errors go out as
  `{"status": "error", "message": ...}` rather than as a traceback.
- [x] Reference files stay under 200 lines (if modified) — none modified.
- [x] `set -euo pipefail` used in any new shell scripts — no shell scripts added.
- [x] CHANGELOG.md updated with the change — an `Unreleased / Fixed` section covering all
  four defects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,33 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Spend ledger lost concurrent writes (`dataforseo_costs.py`).** The ledger took its
+  file lock twice, once to read and once to write, and released it in between, so two
+  concurrent `log` calls both read the same ledger and the second silently overwrote the
+  first. Measured 3 entries lost out of 20 concurrent writes, with 1 to 3 lost on every
+  repeat run. This is not platform-specific. It matters because `seo-audit` fans out to
+  up to 15 parallel specialists, three of which call DataForSEO and each of which logs
+  after every call, so the ledger under-reported most during the runs that spend the
+  most. One exclusive lock now spans the entire read-modify-write in both `log` and
+  `reset`.
+- **No file locking at all on Windows.** The module set `fcntl = None` when the import
+  failed and then took every unlocked code path, so Windows ran the ledger with no
+  locking whatsoever. Locking now falls back to `msvcrt`, and when neither primitive is
+  available the operation fails loudly rather than proceeding unlocked.
+- **A corrupt ledger silently reset the budget to zero.** The non-atomic
+  `open(..., "w")` write could leave a truncated file if the process died mid-write, and
+  the reader caught `JSONDecodeError` and returned an empty ledger, which the next write
+  then persisted, discarding the whole spend history. Ledger writes are now atomic
+  (tempfile plus `os.replace`, matching the idiom already used in `google_auth.py`), and
+  an unparseable ledger fails closed with the file left in place for inspection.
+- **`CLAUDE_SEO_CONFIG_DIR`** now overrides the config and ledger location, so the new
+  `tests/test_dataforseo_costs.py` concurrency and durability tests run against an
+  isolated ledger instead of the operator's real spend history.
+
 ## [2.2.4] - 2026-07-20
 
 Community maintenance release following a full review of every open issue and pull request.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **`CLAUDE_SEO_CONFIG_DIR`** now overrides the config and ledger location, so the new
   `tests/test_dataforseo_costs.py` concurrency and durability tests run against an
   isolated ledger instead of the operator's real spend history.
+- **The Banana cost ledger had the same defect, with no locking at all.**
+  `extensions/banana/scripts/cost_tracker.py` read and wrote `~/.banana/costs.json` with
+  no file locking on any platform and a non-atomic write. Measured 4 entries lost out of
+  20 concurrent `log` calls, and 3 of those processes crashed outright with
+  `JSONDecodeError` after reading a half-written file. It is also harder to notice than
+  the DataForSEO case: this ledger carries running aggregates (`total_cost`,
+  `total_images`, `daily`) incremented in the same write as the entry, so a lost write
+  drops both together and the file still looks internally consistent. Given the same
+  treatment: one exclusive lock across the whole read-modify-write in `log`, atomic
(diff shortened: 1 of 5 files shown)


annairdrychilltechindiapvtltd-source (approved):

AgriciDaniel:
Landed in v2.2.6 as commit 227a00 and b100ee (cherry-picked with -x, your authorship preserved). Thank you.

PR #197, opened 2026-07-29
Add MiniMax image generation to the Banana REST fallback

octo-patch:
Reason: The Banana direct REST fallback could not target the MiniMax image_generation API, so text-to-image generation had no MiniMax path.

## What changed

- `extensions/banana/scripts/generate.py`: add an opt-in `--provider minimax` path to the stdlib-only fallback generator, leaving the existing default path unchanged.
  - Global (`https://api.minimax.io/v1/image_generation`) and CN (`https://api.minimaxi.com/v1/image_generation`) endpoints, selected with `--region {global_en,cn_zh}`.
  - `image-01` (default) and `image-01-live` models.
  - Request controls: `aspect_ratio` or explicit `--width`/`--height`, `--response-format {url,base64}`, `--seed`, `--count` (n), `--prompt-optimizer`/`--no-prompt-optimizer`, and `--subject-reference` (URL, data URI, or local file encoded to a base64 data URI).
  - Bearer authorization; response parsing for both `data.image_urls` (downloaded and saved to the existing output directory) and base64 payloads (decoded and saved); `base_resp.status_code` error surfacing and `metadata` success/failed counts.
  - API-key redaction on error output.
- `tests/test_banana_minimax_image.py`: unit tests for endpoint/model configuration, request-body assembly, URL and base64 response parsing, subject-reference encoding, the base64 save flow, and key redaction.

## Checks

- `python3 -m py_compile extensions/banana/scripts/generate.py`
- `python3 -m pytest tests/test_banana_minimax_image.py tests/test_banana_api_key_safety.py tests/test_banana_install_layout.py -q` -> 14 passed

PR #203, opened 2026-08-04
Add MiniMax image-to-image editing to the Banana REST fallback

octo-patch:
Reason: The Banana image-to-image edit fallback had no MiniMax image_generation path, so image-to-image edits could not target the MiniMax API.

## What changed

- `extensions/banana/scripts/edit.py`: add an opt-in `--provider minimax` path to the stdlib-only fallback editor, leaving the existing default path unchanged.
  - Global (`https://api.minimax.io/v1/image_generation`) and CN (`https://api.minimaxi.com/v1/image_generation`) endpoints, selected with `--region {global_en,cn_zh}`.
  - `image-01` (default) and `image-01-live` models.
  - The input `--image` is sent as a subject reference (local files encoded to base64 data URIs); repeatable `--subject-reference` adds more references.
  - Request controls: `--aspect-ratio` or explicit `--width`/`--height`, `--response-format {url,base64}`, `--seed`, `--count` (n), and `--prompt-optimizer`/`--no-prompt-optimizer`.
  - Bearer authorization; response parsing for both `data.image_urls` (downloaded and saved to the existing output directory) and base64 payloads (decoded and saved); `base_resp.status_code` error surfacing and `metadata` success/failed counts.
  - API-key redaction on error output.
- `tests/test_banana_minimax_edit.py`: unit tests for endpoint/model configuration, request-body assembly, subject-reference encoding, URL and base64 response parsing, the save flow, and key redaction.

## Checks

- `python3 -m py_compile extensions/banana/scripts/edit.py`
- `python3 -m pytest tests/test_banana_minimax_edit.py tests/test_banana_api_key_safety.py tests/test_banana_install_layout.py -q` -> 15 passed
- `python3 -m pytest tests/ -q` -> 419 passed

### DIFF
--- a/extensions/banana/scripts/edit.py
+++ b/extensions/banana/scripts/edit.py
@@ -7,6 +7,13 @@ Uses only Python stdlib (no pip dependencies).
 Usage:
     edit.py --image path/to/image.png --prompt "remove the background"
             [--model MODEL] [--api-key KEY]
+
+    edit.py --provider minimax --image path/to/image.png --prompt "make it sunset"
+            [--region global_en|cn_zh] [--model image-01|image-01-live]
+            [--aspect-ratio 16:9 | --width W --height H]
+            [--response-format url|base64] [--seed N] [--count N]
+            [--prompt-optimizer|--no-prompt-optimizer]
+            [--subject-reference SRC] [--api-key KEY]
 """
 
 import argparse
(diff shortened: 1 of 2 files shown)


octo-patch:
I pushed a small follow-up commit to align the new MiniMax edit test with the repository Ruff formatting. I ran the 15 related tests, Ruff, Python bytecode compilation, and git diff checking.

AgriciDaniel:
Same two notes as #197, since `edit.py` carries the same pattern.

**1.** `edit.py:291` does `urllib.request.urlopen(item, timeout=120)` on a URL taken from the MiniMax response and writes the result to disk, with no `url_safety` validation (the file doesn't import it). CLAUDE.md requires outbound URLs to go through `scripts/url_safety.py` — could this use `safe_requests_get()`?

**2.** The subject-reference helper reads an arbitrary local path and base64-encodes it into the outbound request. Worth constraining to an allowlisted directory.

Happy to re-review once #197 settles, since the fix should be the same shape in both.

---

octo-patch:
Thanks for the review. I addressed both notes in `extensions/banana/scripts/edit.py` and pushed a new commit. (1) The result-image download now goes through `url_safety.safe_requests_get()` instead of a raw `urlopen`, so it gets the canonical pre-flight validation and DNS pinning; if that module or its `requests` dependency is unavailable the URL download fails closed rather than silently fetching unvalidated, and `--response-format base64` still needs no network fetch. (2) Local image inputs are resolved first (so symlinks cannot escape) and must sit under the output directory, the current working directory, or a directory listed in `NANOBANANA_INPUT_DIRS`, and must carry a supported image suffix. Checks: full `pytest tests/` (425 passed), `ruff check` on both changed files, and the repo py_compile sweep over all tracked Python files.

AgriciDaniel:
Declining for now together with #197 (see the comment there for the reopen conditions). Thank you for the careful offline test coverage.

From the 2026-09-10 full review and the Phase 3 decisions; the conditions above are what would reopen this.

PR #204, opened 2026-08-04
fix(extensions): write MCP config to ~/.claude.json, not settings.json

juliandickie:
## The bug

The `dataforseo`, `firecrawl`, `ahrefs` and `banana` installers merge their MCP server block into `~/.claude/settings.json`.

Claude Code does not read `mcpServers` from that file. The block is written, the installer prints its success message, and the server never loads — `claude mcp list` does not show it and no tools appear. Reinstalling cannot fix it, because the destination itself is wrong.

MCP servers belong in `~/.claude.json`, which is what `claude mcp add` writes.

## Why it goes unnoticed

Nothing errors. The JSON is valid, the key name looks right, the file is the one most people associate with Claude Code configuration, and the installer reports success. The failure only shows up later as "the extension installed but there are no tools", which reads like a plugin problem rather than a path problem.

## The change

Repoint those four extensions at `~/.claude.json` across install and uninstall scripts (`.sh` and `.ps1`), banana's Python helpers, and the setup docs.

The existing atomic read-modify-write is kept as-is, so unrelated keys in that file are preserved — this matters more than it did for `settings.json`, since `~/.claude.json` also holds project and session state.

Path variables are renamed (`SETTINGS_FILE` → `MCP_CONFIG_FILE`, and the PowerShell equivalents) so the name matches the file it points at, and each write site gets a short comment recording the distinction so it is not lost again.

## Deliberately not changed

`bing-webmaster`, `profound` and `seranking` also write to `settings.json`, but they write the **`env`** key, which *is* supported there. Moving those would break three working extensions. Only the four that write `mcpServers` are touched.

## Verification

Each installer and uninstaller was executed against a sandboxed `HOME` seeded with an unrelated MCP server and an unrelated top-level key:

| Extension | install | uninstall | servers after round trip | unrelated key | `settings.json` created |
|---|---|---|---|---|---|
| dataforseo | 0 | 0 | `pre-existing` only | preserved | no |
| firecrawl | 0 | 0 | `pre-existing` only | preserved | no |
| banana | 0 | 0 | `pre-existing` only | preserved | no |
| ahrefs | 0 | 0 | `pre-existing` only | preserved | no |

Each wrote only its own server on install, removed only its own on uninstall, and left everything else untouched.

Test suite: **410 passed** on `main` before the change and **410 passed** after.

Also fixed: `ahrefs/install.ps1` resolved the old path through `Join-Path` rather than a literal, so a pattern-based search misses it.

## Known limitation

The PowerShell scripts are checked for path and variable consistency, including a definition/use audit — PowerShell has no `set -u`, so an undefined variable would silently become `$null` rather than failing loudly. They were **not executed**; no PowerShell was available on the machine used. The `.sh` and `.py` paths are functionally tested as above.

### DIFF
--- a/extensions/ahrefs/docs/AHREFS-SETUP.md
+++ b/extensions/ahrefs/docs/AHREFS-SETUP.md
@@ -18,7 +18,7 @@ The installer:
 3. Pre-warms the `@ahrefs/mcp@0.0.11` npm package via `npx --yes` so the first
    MCP call doesn't spend 10+ seconds downloading.
 4. Copies `skills/seo-ahrefs/SKILL.md` into `~/.claude/skills/seo-ahrefs/`.
-5. Atomically writes `mcpServers.ahrefs` into `~/.claude/settings.json`
+5. Atomically writes `mcpServers.ahrefs` into `~/.claude.json`
    with your token in the `env` block. The settings file is `chmod 0o600`
    after the merge (same hardening as the OAuth token).
 
@@ -41,7 +41,7 @@ Re-run the installer to pre-warm or run `npx --yes --package=@ahrefs/mcp@0.0.11 
 
 The Python merge script is idempotent — re-running only replaces the
 `mcpServers.ahrefs.env.AHREFS_API_TOKEN` value, leaving the rest of
-`settings.json` intact.
+`~/.claude.json` intact.
 
 ## Uninstall
 
--- a/extensions/ahrefs/install.ps1
+++ b/extensions/ahrefs/install.ps1
@@ -14,7 +14,10 @@ if (-not (Test-Cmd python)) { throw "Python 3 is required." }
 if (-not (Test-Cmd npx))    { throw "Node 18+ / npx is required." }
 
 $SkillDir = Join-Path $HOME ".claude/skills"
-$SettingsJson = Join-Path $HOME ".claude/settings.json"
+# MCP servers live in ~/.claude.json (the file `claude mcp add` writes).
+# NOT ~/.claude/settings.json - `mcpServers` is not a key Claude Code reads
+# there, so entries written to settings.json silently never load.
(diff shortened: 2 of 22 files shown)


AgriciDaniel:
The premise checks out. Per the Claude Code docs, MCP servers live in `~/.claude.json` (local and user scope) or a project `.mcp.json`; `mcpServers` is not read from `~/.claude/settings.json`. So the current installers write config that silently never loads. Real bug, right destination.

The thing I'd want tightened before merge is that **the blast radius changes**. `~/.claude.json` is the user's primary state file — every project's history — not a small settings file. Two write paths become risky at that size:

- `extensions/firecrawl/install.ps1` uses `ConvertTo-Json -Depth 10`. PowerShell silently truncates beyond `-Depth`, and `~/.claude.json` routinely nests deeper than 10. That can mangle the user's entire config.
- `extensions/banana/scripts/setup_mcp.py` `save_settings()` rewrites the whole file non-atomically, so a crash mid-write loses everything.

The bash installers are fine — they already do argv-passed, atomic, `0600` writes.

Raising the PowerShell depth (or round-tripping without re-serialising the whole document) and making `setup_mcp.py` atomic would make this safe to land. Worth getting right rather than fast, given what the file holds.

---

AgriciDaniel:
The destination correction is right and 3 of 4 installers now write atomically. Two writers still overwrite in place: extensions/firecrawl/install.ps1 (ConvertTo-Json -Depth 10 then Set-Content) and extensions/banana/scripts/setup_mcp.py, and both are reachable entry points. Once those use the same tempfile-plus-rename pattern (and the PowerShell one serialises with enough depth to round-trip the existing file), this can land.

AgriciDaniel:
Landed in v2.3.0 as commit 0dd0538; the firecrawl PowerShell installer and the Banana setup helper now write ~/.claude.json atomically. Thank you.

PR #209, opened 2026-08-08
Fix scripts/ path resolution and interpreter for direct skill/agent invocation

### LINKED ISSUE #208: [Bug]: Sub-skills/agents reference scripts/*.py with no way to locate the plugin root or correct interpreter
### Operating System

macOS 26.4.1 (Darwin 25.4.0)

### Python Version

System `python3 --version` → 3.14.3. This turns out to be part of the bug (see below) — the plugin ships its own `.venv` (Python 3.13) precisely because system Python lacks required dependencies, but nothing in the affected files says to use it.

### Command that failed

`/claude-seo:seo-audit <url>`, and more broadly **any of the 18 sub-agents invoked directly** (e.g. via Claude Code's Task/Agent tool with `subagent_type: claude-seo:seo-performance`, `claude-seo:seo-technical`, etc.) rather than through the top-level `/seo` router.

### Full error output

Running the full audit, several sub-agents either silently degraded or explicitly gave up. Representative excerpt from the `seo-performance` sub-agent's own report:

```
The workflow's designated tools (scripts/pagespeed_check.py, scripts/render_page.py,
skills/seo/references/cwv-thresholds.md) do not exist anywhere on this filesystem —
they were never installed in this project.
```

That conclusion is wrong — all three exist — but the agent had no way to find them. Manually reproducing the same call:

```
$ python3 scripts/render_page.py --help
python3: can't open file '.../skills/seo-audit/scripts/render_page.py': [Errno 2] No such file or directory
```

and, once the correct path is found by hand:

```
$ python3 scripts/render_page.py --help
Traceback (most recent call last):
  File ".../scripts/url_safety.py", line 84, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

### Expected behavior

Any skill or sub-agent file that references `scripts/*.py` should be runnable standalone — without the invoking agent having to independently rediscover the plugin's own install path or its bundled virtualenv.

### Root cause (two separate, compounding bugs)

1. **No reliable way to locate the plugin root.** `skills/seo/SKILL.md` (the top-level router) states "Scripts: Located at the plugin root `scripts/` directory" — but that line is never propagated into the files that actually get invoked directly. `skills/seo-audit/SKILL.md` itself contains only `SKILL.md` and `LICENSE.txt` — no `scripts/` folder — because the real `scripts/` directory lives once at the plugin root, shared across all skills/agents. I found **21 files** (11 in `skills/*/SKILL.md`, 10 in `agents/*.md`) that reference `scripts/*.py` with a bare relative path and no note about where it resolves from. Only 3 files (`seo-cluster/SKILL.md`, `seo-schema.md`, `seo-visual.md`) mention "the plugin root" at all, and even those don't say *how* to find it — which is presumably why the `seo-schema` sub-agent had to spend several tool calls searching the filesystem before it found its own scripts, when it should have been immediate.

2. **Even once located, `scripts/` needs the bundled `.venv`, not system Python.** `render_page.py` imports `requests` (via `url_safety.py`), which isn't installed under system `python3` in this environment. The plugin ships `.venv/bin/python3` with the right dependencies, but no affected file mentions it.

Net effect: an agent invoking a sub-skill/sub-agent directly (the normal way Claude Code's own multi-agent delegation works) has no reliable path to the plugin's actual tooling, and silently falls back to weaker methods (raw `curl`/`WebFetch`) or reports the scripts as missing outright — which is a materially worse audit (e.g. no real Core Web Vitals data, `pagespeed_check.py` never runs).

### Suggested fix (patch available)

I fixed this locally by inserting a short, self-locating snippet right after the YAML frontmatter in all 21 affected files:

```bash
PLUGIN_ROOT=$(dirname "$(find ~/.claude/plugins/cache -maxdepth 4 -type d -path '*/claude-seo/*/scripts' 2>/dev/null | grep -v '/extensions/' | head -1)")
PY="$PLUGIN_ROOT/.venv/bin/python3"
"$PY" "$PLUGIN_ROOT/scripts/<name>.py" ...
```

Two things worth flagging about the resolution command itself, since a naive version breaks in a way that isn't obvious until tested:

- It must anchor to `~/.claude/plugins/cache` specifically — `~/.claude/plugins/marketplaces/AgriciDaniel-claude-seo/` also contains a `scripts/` directory (it's the marketplace source checkout) but has **no** `.venv`, so a broader search can non-deterministically resolve to the wrong copy depending on filesystem traversal order.
- It needs to exclude `extensions/*/scripts` (e.g. `extensions/banana/scripts`), which would otherwise also match.

I verified the fix end-to-end (not just that it edits cleanly) by running `render_page.py`, `pagespeed_check.py`, `google_auth.py`, and `backlinks_auth.py` through the resolved path/interpreter — all four execute correctly now.

I'm happy to open a PR with the diff (21 files, same 8-line insertion in each, or you may prefer a shared constants file / a single `scripts/_bootstrap.sh` that every SKILL.md and agent sources instead of repeating the snippet — that'd be a cleaner long-term fix than what I did locally). Let me know which you'd prefer and I'll put the PR together.

vickaul-ai:
## Summary

Fixes #208.

21 files (11 `skills/*/SKILL.md`, 10 `agents/*.md`) reference `scripts/*.py` with a bare relative path and no indication of where it resolves from — only correct when the caller already has the plugin root as its working directory, which isn't the case when a skill or agent is invoked directly (e.g. via Claude Code's Task/Agent tool, or `/claude-seo:seo-audit` itself) rather than through the `/seo` router. Only 3 files mentioned "the plugin root" at all, and none said how to locate it.

Also fixes a second, compounding bug: system `python3` lacks dependencies the scripts need (e.g. `requests`, imported via `url_safety.py`) — the plugin ships its own `.venv` for exactly this reason, but nothing in the affected files said to use it.

## Fix

Inserts a short, self-locating snippet right after the YAML frontmatter in each of the 21 files:

```bash
PLUGIN_ROOT=$(dirname "$(find ~/.claude/plugins/cache -maxdepth 4 -type d -path '*/claude-seo/*/scripts' 2>/dev/null | grep -v '/extensions/' | head -1)")
PY="$PLUGIN_ROOT/.venv/bin/python3"
"$PY" "$PLUGIN_ROOT/scripts/<name>.py" ...
```

Two non-obvious things the resolution command specifically guards against (a naive version breaks on both, only visible once tested against a real install):

- It anchors to `~/.claude/plugins/cache` specifically — `~/.claude/plugins/marketplaces/<owner>-claude-seo/` also has a `scripts/` directory (the marketplace source checkout) but no `.venv`, so a broader search can non-deterministically resolve to the wrong copy.
- It excludes `extensions/*/scripts` (e.g. `extensions/banana/scripts`), which would otherwise also match.

## Test plan

- [x] Confirmed all 21 target files exist at these paths and matched the same broken pattern before the fix
- [x] Verified end-to-end against a real install: ran `render_page.py`, `pagespeed_check.py`, `google_auth.py`, and `backlinks_auth.py` through the exact resolution steps now written into the files — all four execute correctly
- [x] Confirmed the resolution command correctly ignores the marketplace-checkout and `extensions/` false-positive matches described above

## Note on scope

This applies the same fix pattern to every currently-affected file rather than introducing a new shared bootstrap mechanism (e.g. a single `scripts/_bootstrap.sh` sourced by each file), to keep the diff minimal and directly verifiable. Happy to follow up with that refactor separately if you'd prefer to consolidate it — flagged as an alternative in the linked issue.

### DIFF
--- a/agents/seo-backlinks.md
+++ b/agents/seo-backlinks.md
@@ -6,6 +6,14 @@ maxTurns: 20
 tools: Read, Bash, Write, Glob, Grep
 ---
 
+**Locating scripts:** every `scripts/*.py` path in this file is relative to the plugin's root directory, not this file's own folder — and needs the plugin's bundled virtualenv interpreter (system `python3` is missing dependencies like `requests`/`playwright`). Resolve both once, then reuse for every script call below:
+
+```bash
+PLUGIN_ROOT=$(dirname "$(find ~/.claude/plugins/cache -maxdepth 4 -type d -path '*/claude-seo/*/scripts' 2>/dev/null | grep -v '/extensions/' | head -1)")
+PY="$PLUGIN_ROOT/.venv/bin/python3"
+"$PY" "$PLUGIN_ROOT/scripts/<name>.py" ...
+```
+
 You are a backlink profile analyst. When delegated tasks during an SEO audit:
 
 1. Check credentials: `claude-seo run backlinks_auth.py --check --json`
--- a/agents/seo-content.md
+++ b/agents/seo-content.md
@@ -6,6 +6,14 @@ maxTurns: 15
 tools: Read, Bash, Write, Grep
 ---
 
+**Locating scripts:** every `scripts/*.py` path in this file is relative to the plugin's root directory, not this file's own folder — and needs the plugin's bundled virtualenv interpreter (system `python3` is missing dependencies like `requests`/`playwright`). Resolve both once, then reuse for every script call below:
+
+```bash
+PLUGIN_ROOT=$(dirname "$(find ~/.claude/plugins/cache -maxdepth 4 -type d -path '*/claude-seo/*/scripts' 2>/dev/null | grep -v '/extensions/' | head -1)")
+PY="$PLUGIN_ROOT/.venv/bin/python3"
+"$PY" "$PLUGIN_ROOT/scripts/<name>.py" ...
+```
+
 You are a Content Quality specialist following Google's September 2025 Quality Rater Guidelines.
 
 When given content to analyze:
--- a/agents/seo-drift.md
+++ b/agents/seo-drift.md
@@ -9,6 +9,14 @@ maxTurns: 15
 tools: Read, Bash, Write, Glob, Grep
 ---
 
+**Locating scripts:** every `scripts/*.py` path in this file is relative to the plugin's root directory, not this file's own folder — and needs the plugin's bundled virtualenv interpreter (system `python3` is missing dependencies like `requests`/`playwright`). Resolve both once, then reuse for every script call below:
+
+```bash
+PLUGIN_ROOT=$(dirname "$(find ~/.claude/plugins/cache -maxdepth 4 -type d -path '*/claude-seo/*/scripts' 2>/dev/null | grep -v '/extensions/' | head -1)")
+PY="$PLUGIN_ROOT/.venv/bin/python3"
+"$PY" "$PLUGIN_ROOT/scripts/<name>.py" ...
+```
+
 <!-- Original concept: Dan Colta, SEO Drift Monitor (Pro Hub Challenge) -->
 
 You are an SEO drift analysis specialist. You detect regressions in on-page SEO
--- a/agents/seo-ecommerce.md
+++ b/agents/seo-ecommerce.md
@@ -9,6 +9,14 @@ maxTurns: 20
 tools: Read, Bash, Write, Glob, Grep
 ---
 
+**Locating scripts:** every `scripts/*.py` path in this file is relative to the plugin's root directory, not this file's own folder — and needs the plugin's bundled virtualenv interpreter (system `python3` is missing dependencies like `requests`/`playwright`). Resolve both once, then reuse for every script call below:
+
+```bash
+PLUGIN_ROOT=$(dirname "$(find ~/.claude/plugins/cache -maxdepth 4 -type d -path '*/claude-seo/*/scripts' 2>/dev/null | grep -v '/extensions/' | head -1)")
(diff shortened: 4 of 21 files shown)


AgriciDaniel:
Thanks for the careful write-up in #208 — the diagnosis is right. I counted 25 `scripts/*.py` references in `agents/`/`skills/` that bypass the wrapper, against 28 that already use `claude-seo run`. That inconsistency is real and worth fixing.

I'd push back on the approach though. `scripts/runtime.py` already solves this, and handles cases this block doesn't:

- it honours `CLAUDE_PLUGIN_DATA` / `CLAUDE_PLUGIN_ROOT` rather than assuming one install layout;
- `_venv_python()` returns `Scripts/python.exe` on Windows, whereas `"$PLUGIN_ROOT/.venv/bin/python3"` is Unix-only;
- `bin/claude-seo` already probes `CLAUDE_SEO_PYTHON`, `py -3`, `python3`, `python` with a 3.10 floor.

The `find ~/.claude/plugins/cache -maxdepth 4 ... | head -1` block also breaks for git-clone and marketplace installs, and it's duplicated into 20 files, which is 20 places to update when the layout changes.

Would you be up for converting those 25 raw references to `claude-seo run` instead? That's the same fix with less surface, and it's what CLAUDE.md already specifies. The bug report stands on its own either way — thanks for filing it.

---

AgriciDaniel:
Closing: v2.2.5 shipped the managed launcher (claude-seo run) that resolves the plugin root and interpreter on every platform, and v2.2.6 moves it to scripts/ for hosted installs. 20 of the 21 files this PR touched already use it. Thank you for pushing on the problem.

PR #211, opened 2026-08-12
Allow auditing a local dev server behind CLAUDE_SEO_ALLOW_LOCAL

PieterJanDevolder:
## Problem

`url_safety` refuses loopback and private addresses unconditionally — no CLI flag, no environment override. The module reads no environment at all today.

That default is correct for a fetcher that follows URLs discovered on crawled pages. But it also means the SEO skills cannot audit a site **before it is deployed**:

```
$ claude-seo run fetch_page.py http://localhost:3000
Error: url_safety: Blocked hostname: localhost
```

For anyone running the skills against a site in development, that is the main workflow — and there is currently no supported way to say "I know, this host is mine". Hosts-file aliasing does not help either, since `validate_url_strict` resolves every A record and rejects the resulting loopback IP.

## Change

An opt-in escape hatch, off unless `CLAUDE_SEO_ALLOW_LOCAL=1`:

- **`_blocked(hostname)`** replaces the three direct `_BLOCKED_HOSTNAMES` membership tests. It forgives only the three loopback names (`localhost`, `ip6-localhost`, `ip6-loopback`), and only when opted in, so cloud metadata hostnames stay blocked in every configuration.
- **`_is_local_dev_ip(ip)`** gates the address check: loopback and RFC1918 only.

The second one is worth a note. `ipaddress` reports `169.254.0.0/16` as **private**, so the obvious `is_loopback or is_private` carve-out silently hands back `169.254.169.254`. Link-local, reserved, multicast and unspecified are therefore excluded explicitly. The test for this caught that exact bug in my first draft, which is why it is in the suite.

Only the literal string `"1"` opts in — `"true"`, `"yes"` and `"0"` do not.

## Default behaviour is unchanged

With the variable unset, every existing code path evaluates exactly as before: `_blocked()` reduces to the old membership test and `_is_local_dev_ip()` is never reached.

## Tests

Six new cases in `tests/test_url_safety.py`:

| Test | Asserts |
|---|---|
| `test_localhost_blocked_by_default` | unset env still blocks, both validators |
| `test_localhost_allowed_when_opted_in` | opted-in path works through `validate_url` **and** `validate_url_strict`, including the resolved-IP pin |
| `test_private_and_loopback_literals_allowed_when_opted_in` | `127.0.0.1`, `192.168.1.10` |
| `test_opt_in_does_not_unblock_cloud_metadata_names` | three metadata hostnames stay blocked while opted in |
| `test_opt_in_does_not_unblock_link_local_metadata_ip` | `169.254.169.254` stays blocked by name and by address |
| `test_only_the_literal_value_1_opts_in` | `""`, `"0"`, `"true"`, `"yes"`, `"TRUE"` do not opt in |

Run on Windows / Python 3.13: **96 passed, 3 failed**. The three failures are the pre-existing `test_{save,load}_oauth_token_*` POSIX-permission cases, which fail identically on a clean checkout of `main` (baseline there: 90 passed, 3 failed).

## Notes

Happy to adjust the variable name, restrict the carve-out to `validate_url_strict` only, or add a line to the docs if you would like this documented alongside `CLAUDE_SEO_PYTHON`.

### DIFF
--- a/scripts/url_safety.py
+++ b/scripts/url_safety.py
@@ -73,6 +73,7 @@ in SECURITY.md.
 from __future__ import annotations
 
 import ipaddress
+import os
 import re
 import socket
 import threading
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
Default policy is unchanged and the metadata negative controls hold, which is the hard part. Two findings block the current shape: with the flag on, getaddrinfo("192.168.1.10") succeeds inside a pinned fetch, so redirects and browser subresources to any RFC 1918 host open too; and validate_url("http://127.0.0.1:3000/") is still False while localhost:3000 is True. A redesign as an explicit allowlist of top-level targets (host plus port), applied only to the initial URL and never to redirects or subresources, would cover the local dev server case and keep the rest fail-closed. Since v2.2.6 also refuses 100.64.0.0/10, the same allowlist should be able to name a Tailscale address.

AgriciDaniel:
Closing with thanks: v2.3.0 ships the redesign discussed in the review as CLAUDE_SEO_LOCAL_TARGETS, an explicit allowlist of top-level targets (host or host:port) that never applies to redirects or subresources and can never name a metadata address. It covers the local dev server case this PR set out to solve, plus Tailscale addresses now that 100.64.0.0/10 is refused. Your tests informed the negative controls.

Issue #188, opened 2026-07-27
validate-schema.py hook: bare "REPLACE" substring blocks edits on valid JSON-LD (e.g. "renal replacement therapy")


mg26984-hash:
## Summary

`hooks/validate-schema.py` treats the bare token `REPLACE` as placeholder text using a
case-insensitive **substring** match. Because "placeholder" is in `critical_keywords`, the match is
classified as critical and the hook exits **2, blocking the edit**.

Any site whose JSON-LD legitimately contains the word *replacement*, *replaces*, or *replaced* is
therefore blocked from editing that file, with a misleading error. Medical sites are hit hardest:
"renal **replace**ment therapy" is standard nephrology vocabulary, and "hormone replacement
therapy", "joint replacement", "replacement parts" (e-commerce) and "INP **replace**s FID" (which
appears in this plugin's own reference material) all trip it.

## Reproduction

`probe.html`:

```html
<!doctype html><html><head>
<script type="application/ld+json">
{"@context":"https://schema.org","@type":"MedicalOrganization","name":"Arab Society of Nephrology and Renal Transplantation","description":"Advancing renal replacement therapy across the Arab world."}
</script>
</head><body></body></html>
```

```
$ python3 hooks/validate-schema.py probe.html
🛑 Schema validation ERRORS (blocking):
  - Block 1: Contains placeholder text: REPLACE
$ echo $?
2
```

The JSON-LD is valid, uses an active type, and contains no placeholder.

## Expected

Exit 0. Genuine placeholder tokens (`REPLACE`, `REPLACE_ME`, `REPLACE THIS URL`) should still
block; ordinary English words that merely start with "replace" should not.

## Cause

`_validate_schema_object()` mixes bracketed placeholders (`[Business Name]`, `[URL]`, …), which are
safe to substring-match, with the unbracketed token `REPLACE`, which is not:

```python
placeholders = [..., "[INSERT", "REPLACE", "[URL]", "[Email]"]
text = json.dumps(obj)
for p in placeholders:
    if p.lower() in text.lower():
        errors.append(f"{prefix}: Contains placeholder text: {p}")
```

## Suggested fix

Keep bracketed tokens on substring matching and move unbracketed tokens to a word-boundary regex
that rejects a following lowercase letter — `REPLACE`, `REPLACE_ME` and `REPLACE THIS` still match,
while `replacement` / `replaces` / `replaced` do not:

```python
[...]

PR #240, opened 2026-08-15
fix(schema): avoid replacement placeholder false positive

### LINKED ISSUE #205: validate-schema.py blocks legitimate edits: bare REPLACE substring matches "replacement"
**Version:** 2.2.4 · **Hook:** `hooks/validate-schema.py` (PostToolUse, `Edit|Write`)

The placeholder check matches the bare token `REPLACE` case-insensitively as a substring:

```python
placeholders = [..., "REPLACE", ...]
text = json.dumps(obj)
for p in placeholders:
    if p.lower() in text.lower():
        errors.append(f"{prefix}: Contains placeholder text: {p}")
```

`"replace"` is a substring of **"replacement"**, and placeholder findings are classified `critical` → `sys.exit(2)`, which **blocks the write**.

**Reproduce:** edit a `.tsx` containing JSON-LD with `"description": "Replacement coils for the Caliburn G2"` → write is blocked with `🛑 Contains placeholder text: REPLACE`.

**Impact:** any ecommerce site selling replacement parts. Every other entry in the list is bracket-delimited (`[Business Name]`, `[INSERT`), so `REPLACE` is the only one that can match real prose. Because the hook is `Edit|Write` with no directory scope, it fires in every project.

**Suggested fix:** require delimiters, e.g. `[REPLACE` / `REPLACE_ME` / `{{REPLACE}}`, or word-boundary match `\bREPLACE\b` on the raw token rather than a case-folded substring.

**Workaround:** we emptied `hooks/hooks.json`, which the version-pinned cache path means must be re-applied after each update.

ousamabenyounes:
## Summary

Prevent the schema validation hook from blocking legitimate JSON-LD text such as
"Replacement coils" or "Replace the cartridge every 6 months". Genuine uppercase
placeholder tokens remain blocking, including underscore forms such as `REPLACE_ME`
and `REPLACE_WITH_URL`.

Fixes #205.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Test verification (RED -> GREEN)

Prior PR implementation, regression tests only:

```text
FAILED test_replace_verb_is_not_treated_as_placeholder: expected 0, got 2
FAILED test_underscore_replace_placeholder_still_blocks: expected 2, got 0
2 failed, 4 passed
```

Patched branch:

```text
Focused policy suite: 6 passed
Full local CI: 414 passed
Upstream baseline local CI: 410 passed
Changed executable production lines: 100% covered
```

## Checklist

- [ ] Tested with a real URL before submitting (not applicable: local hook validation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ patches excluded. No breaking command changes. Full suite: 410 passing tests.
 
 ### Fixed
 
+- Schema validation no longer treats words such as "replacement" as unresolved `REPLACE`
+  placeholders, while standalone placeholder tokens remain blocked.
 - GSC queries now honor exact result limits, validate dimensions before API access, support
   dimensionless totals, and report whether aggregate totals are complete.
 - Bing Webmaster backlink commands now use supported endpoints, bounded pagination, deduplication,
--- a/hooks/validate-schema.py
+++ b/hooks/validate-schema.py
@@ -36,6 +36,8 @@ import re
 import sys
 from typing import List
 
+BARE_PLACEHOLDER = "REPLACE"
+
 
 def validate_jsonld(content: str) -> List[str]:
     """Validate JSON-LD blocks in HTML content."""
@@ -87,7 +89,6 @@ def _validate_schema_object(obj: dict, block_num: int) -> List[str]:
         "[Address]",
         "[Your",
         "[INSERT",
-        "REPLACE",
         "[URL]",
         "[Email]",
     ]
@@ -95,6 +96,8 @@ def _validate_schema_object(obj: dict, block_num: int) -> List[str]:
     for p in placeholders:
         if p.lower() in text.lower():
             errors.append(f"{prefix}: Contains placeholder text: {p}")
+    if re.search(rf"\b{re.escape(BARE_PLACEHOLDER)}\b", text, re.IGNORECASE):
+        errors.append(f"{prefix}: Contains placeholder text: {BARE_PLACEHOLDER}")
 
     # Check for deprecated types
     schema_type = obj.get("@type", "")
--- a/hooks/validate-schema.py
+++ b/hooks/validate-schema.py
@@ -37,6 +37,9 @@ import sys
 from typing import List
 
 BARE_PLACEHOLDER = "REPLACE"
+BARE_PLACEHOLDER_PATTERN = re.compile(
+    rf"\b{re.escape(BARE_PLACEHOLDER)}(?:_[A-Z]+)*\b"
+)
 
 
 def validate_jsonld(content: str) -> List[str]:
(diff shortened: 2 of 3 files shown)


AgriciDaniel:
Verified: the false positive reproduces on `main` — `"renal replacement therapy"` yields `Contains placeholder text: REPLACE` — and this branch's tests pass (`4 passed`). The word-boundary regex is the right fix, and it covers both #205 and #188 (which are duplicates of each other).

One sequencing note: this conflicts with your #241, since both rewrite `_validate_schema_object()` — `hooks/validate-schema.py`, `tests/test_schema_hook_policy.py` and `CHANGELOG.md` all collide. My suggestion is to land #241 first (it changes the function signature) and rebase this one on top, which should be the smaller conflict. Either order works as long as the second gets a rebase.

---

AgriciDaniel:
Following up — I probed the new pattern against a wider set of inputs, and I think `\bREPLACE\b` with `re.IGNORECASE` trades one false positive for two problems:

```
'Replace the cartridge every 6 months' -> BLOCKED   <- legitimate product copy
'Easy to replace filter'               -> BLOCKED   <- legitimate product copy
'Do not replace under warranty'        -> BLOCKED   <- legitimate product copy
'REPLACE_ME'                           -> allowed   <- real placeholder, now missed
'REPLACE_WITH_URL'                     -> allowed   <- real placeholder, now missed
```

Two things going on:

1. **`_` is a word character**, so `\bREPLACE\b` doesn't match `REPLACE_ME` or `REPLACE_WITH_URL` — arguably the most common placeholder forms — and they now pass the hook.
2. **`IGNORECASE` catches the ordinary English verb.** "Replace the cartridge every 6 months" is exactly the kind of sentence that belongs in a `Product` description, and it would now block the edit. For a `Product` schema that's likely to fire more often than the original `"replacement"` bug did.

Making the match all-caps, underscore-aware and case-**sensitive** handles every case I threw at it:

```python
re.search(r"\bREPLACE(_[A-Z]+)*\b", text)   # no IGNORECASE
```

```
'REPLACE'                              -> BLOCKED
'REPLACE_ME'                           -> BLOCKED
'REPLACE_WITH_URL'                     -> BLOCKED
'Replace the cartridge every 6 months' -> allowed
'Easy to replace filter'               -> allowed
'renal replacement therapy'            -> allowed
'Replacement coils'                    -> allowed
```

The rationale being that a genuine unresolved placeholder is a screaming-caps token, whereas lowercase "replace" is just English. Worth adding the `REPLACE_ME` and "Replace the cartridge..." cases to the tests either way — they're the ones that would catch a future regression here.

Your diagnosis of the original bug is still right, and #205/#188 are real. This is just about where the new boundary sits.

---

AgriciDaniel:
Thank you for identifying the false positive and helping refine the matching boundary. #261 shipped the case-sensitive uppercase-token solution, preserving `REPLACE` and `REPLACE_TITLE` blocking while allowing ordinary “replace” and “replacement” prose. Released in v2.2.5, so this PR is now superseded.

Issue #186, opened 2026-07-26
validate-schema.py hook crashes with UnicodeEncodeError on Windows (cp1252)


ortorey1976-svg:
## Summary

The `PostToolUse` schema-validation hook crashes with `UnicodeEncodeError` on Windows whenever it has something to report. The emoji in the output (`⚠️` / `🛑`) cannot be encoded by the default `cp1252` console codec, so the hook dies with a traceback instead of printing its findings.

This means **the hook never delivers a warning or a blocking error on Windows** — it only ever produces a Python traceback.

## Environment

| | |
|---|---|
| claude-seo | 2.2.4 (plugin install) |
| OS | Windows 11 (26200) |
| Python | 3.14.3 (`C:\Python314\python.exe`) |
| Node | v24.14.0 |
| Console codec | cp1252 |

## Reproduction

```bash
cat > /tmp/repro.html <<'HTML'
<html><head><script type="application/ld+json">
{"@context":"https://schema.org","@graph":[
 {"@type":"MedicalBusiness","@id":"https://ex.com/#b","name":"X"}]}
</script></head><body>ok</body></html>
HTML

py -3 hooks/validate-schema.py /tmp/repro.html
```

## Actual

```
Traceback (most recent call last):
  File "...\hooks\validate-schema.py", line 197, in <module>
    main()
  File "...\hooks\validate-schema.py", line 183, in main
    print("\u26a0\ufe0f  Schema validation warnings:")
  File "C:\Python314\Lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-1: character maps to <undefined>
```

## Expected

```
⚠️  Schema validation warnings:
  - Block 1: Missing @type
```

Setting `PYTHONIOENCODING=utf-8` produces the expected output, which confirms the cause.

## Note on 2.2.4 changelog

The 2.2.4 notes mention *"hook and child-process output is UTF-8"*. That fix does not appear to cover this path — `hooks/run-python-hook.js` spawns with `stdio: "inherit"` and does not set `PYTHONIOENCODING`, so the child inherits the Windows console codec.

## Suggested fixes

Either would work; the first is the smallest:

1. In `hooks/run-python-hook.js`, set the encoding for the child process:
   ```js
   const result = spawnSync(candidate.exe, [...candidate.args, hookScript, ...hookArgs], {
[...]

PR #243, opened 2026-08-15
Fix three silent-failure bugs (#210, #186, #208), add a Python syntax gate, and triage all open PRs/issues

AgriciDaniel:
## Summary

A skeptical review of all 26 open PRs and 12 open issues, plus fixes for the confirmed bugs that had no working PR behind them.

Every claim here was verified against the code rather than the description: PR branches were fetched and their own tests executed, merges tested pairwise, and external claims checked against primary sources.

**The connecting theme is silent failure.** In each case the tool reported success while doing less than it claimed — the failure mode that is hardest to notice and most corrosive to trust in an auditing tool.

### Fixes

- **`agent_ux_check.py` scored empty documents 90/100 (#210).** `audit()` returned early only when `render_page()` reported an error, so a successful fetch returning an empty body flowed into `score()`. Every signal `analyze_html()` counts collapses to zero on an empty document, which read as a clean page. Now reported as a render error.
- **`agent_ux_check.py` inflated scores when the a11y tree was missing (#210, Bug 2).** The snapshot is best-effort and failures are swallowed to `None`, after which `score()` skips every accessibility deduction: the same page scores **75** with a tree and **100** without one, with nothing indicating half the analysis never ran. Now reports `partial: true` plus an explicit issue, and the CLI prints `(partial)`.
  - Note: the issue's literal claim that the tree is "never captured" is inaccurate — it *is* captured at `render_page.py:441` under the mode `audit()` uses. The defect is the silent degradation beside it.
- **`validate-schema.py` failed open on Windows (#186).** Encoding the banner markers to cp1252 raised `UnicodeEncodeError` *before* `sys.exit(2)` ran, so the hook exited **1** — warnings-only. A blocking schema finding was therefore silently non-blocking on any cp1252 console. Only the error handler is relaxed; forcing UTF-8 onto such a console would produce mojibake instead.

### Added

- **`compile-check` CI job** covering every tracked `.py` file. This **widens** existing coverage rather than adding something missing: the current `lint` job already compiles `git ls-files 'scripts/*.py'` (53 files). This one covers 96 — adding **35 under `tests/`, 7 under `extensions/banana/scripts/`, and `hooks/validate-schema.py`**, none of which were checked before. Both files this PR edits outside `scripts/` fall in that gap, as do the files #197/#203 touch. Dependency-free, so it cannot fail for environment reasons.

### Changed

- **20 instructional script references across 13 files now use `claude-seo run` (#208).** Applied as explicit one-occurrence-each replacements, not a pattern rewrite: most of the 25 references are prose, and a blind substitution is precisely what left #198 unparseable. Five descriptive references that explain internals are deliberately unchanged.

### Docs

- `docs/REVIEW-open-prs-issues-2026-08.md` — per-PR and per-issue verdicts with reproducible evidence, a conflict map, and a suggested merge order.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [x] Documentation update
- [ ] Refactor / code quality
- [x] Other — CI coverage (Python syntax gate)

## Checklist

- [ ] Tested with a real URL before submitting — **not done, and stated plainly:** outbound network is restricted in this environment. All behaviour is covered by unit tests with mocked fetches instead, and the three suite failures below are network-bound in exactly this way.
- [x] SKILL.md files stay under 500 lines — 219 / 254 / 286 for the three modified.
- [x] Python scripts output JSON — `agent_ux_check.py --json` unchanged, with `partial` added to the payload.
- [x] Reference files stay under 200 lines — 119 and 107 for two of the three. `thinking-framework.md` is **238**, but it is 238 on `main` too; my edits replace 3 lines for a net change of zero. Pre-existing, flagged rather than silently inherited.
- [x] `set -euo pipefail` used in any new shell scripts — the new CI step uses `set -uo pipefail` deliberately, **without `-e`**: it collects every failing file to report them together, which `-e` would abort on the first one.
- [x] CHANGELOG.md updated with the change.

## Verification

Full suite, nothing excluded, with `main` as the control:

| | Passed | Failed |
|---|---|---|
| `main` | 407 | 3 |
| this branch | **414** | 3 |

Exactly **+7 tests, all passing, zero new failures.** The 3 failures are byte-identical on both branches and network-bound (an `example.com` fetch; two `sync_flow` tests receiving HTTP 403 through the sandbox proxy).

Each new regression test was verified to **fail without its corresponding fix** — a test that passes either way proves nothing.

Also clean: `consistency_check.py` (0 errors), `portability_check.py` (0 errors, 0 warnings), and the new syntax gate.

## Correction, and a finding that outranks this PR

An earlier version of this description claimed nothing in CI caught unparseable Python. **That was wrong** — the `lint` job would have caught #198's `SyntaxError`, since `scripts/bing_webmaster.py` matches its glob. Corrected above.

Investigating why it didn't surfaced something more important:

```
PR #198 check runs: 0
PR #242 check runs: 0
PR #182 check runs: 0
```

**No fork-based PR in this repo is running CI at all**, while this same-repo PR got its four checks within seconds. That is consistent with GitHub's "require approval for all external contributors" setting for fork workflow runs. It explains how both #198's `SyntaxError` and #242's failing tests reached review looking clean, and it means external contributors currently get no automated feedback whatsoever.

Worth checking Settings → Actions → Fork pull request workflows. That change would likely do more for contribution quality than anything in this PR.

## Notes for the reviewer

- **This PR is how the new job gets exercised at all.** Workflows trigger only on `push: [main]` and `pull_request: [main]`, so a feature branch never runs CI on its own.
- **No version bump.** `plugin.json` stays at `2.2.4` and the changelog entries sit under `[Unreleased]`, since CLAUDE.md states `main` reflects released history — whether this becomes `2.2.5` is a release decision, not mine.
- **The `aimh` remote is not configured in this environment**, so the documented private-remote-first promote sequence could not be followed from here.
- The one `consistency_check` warning is the review doc itself being unreferenced from the docs tree.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,3 +78,34 @@ jobs:
             exit 1
           fi
           echo "Secret scan clean: no credentials detected in tracked files."
+
+  compile-check:
+    name: Python syntax
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up Python 3.10
+        uses: actions/setup-python@v7
+        with:
+          python-version: "3.10"
+
+      - name: Compile every tracked Python file
+        # Cheapest possible guard against a tracked .py file that does not
+        # parse. pytest does not cover this: an unparseable module surfaces as a
+        # collection error for its own test file, and a script with no test
+        # never gets imported at all. Deliberately dependency-free -- syntax is
+        # checked without installing requirements.txt, so this stays fast and
+        # cannot fail for environment reasons.
+        run: |
+          set -uo pipefail
+          BROKEN=""
+          while IFS= read -r f; do
+            python -m py_compile "$f" 2>/dev/null || BROKEN="${BROKEN}${f}"$'\n'
+          done < <(git ls-files '*.py')
+          if [ -n "$BROKEN" ]; then
+            echo "::error::Tracked Python files that do not parse:"
+            echo "$BROKEN"
+            exit 1
+          fi
+          echo "Syntax check clean: every tracked .py file parses."
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   whitespace-only document. A successful fetch that returned no HTML previously
   scored 90/100, because every signal it counts collapses to zero on an empty
   document; it is now reported as a render error (#210).
+- `render_page.py --json` no longer truncates `extracted_text`. That field is
+  trafilatura's boilerplate-stripped text and the agent instructions point every
+  scoring path at it (E-E-A-T, thin content, word count, citability), but it was
+  capped at 500 characters — a 2000-word page arrived as 73 words, a 27x
+  undercount, so every real page looked thin. Raw HTML fields stay capped so the
+  CLI is still usable over stdio, and a new `truncation` map in the payload
+  records which fields were cut, so a short page is distinguishable from a cut
+  one. `--max-text` caps `extracted_text` on request (#246).
+- `render_page.py --output` now composes with `--json`. The JSON branch exited
+  before the output block ran, so `--json --output` silently wrote nothing —
+  which also broke the documented "use `--output` when you need the full
+  document" escape hatch (#250).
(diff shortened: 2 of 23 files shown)


AgriciDaniel:
Closing as split. Commit 387f7e9 landed in v2.2.6 as 0cea1d; the remaining 22 files were superseded by v2.2.5.

PR #257, opened 2026-08-24
fix: stop flagging SSR/SSG framework pages as SPAs in render_page

stealthdigital:
## Summary

`render_page.py` treated any hydration marker (`<div id="__next">`, `<div id="__nuxt">`, `data-svelte-h=`, `<astro-island `) as proof of a client-rendered SPA and forced headless rendering. Those frameworks routinely serve content-complete raw HTML in SSR/SSG mode (Astro islands pages often have byte-identical raw vs rendered text), so static sites were misreported as `is_spa: true` in audit output and paid an unnecessary Playwright render on every fetch.

Detection is now two-tier:

- **Empty-shell signatures** (bare `<div id="root"></div>` / `<div id="app"></div>` mounts, "enable JavaScript" noscript walls) still flip `is_spa` unconditionally: the raw HTML provably cannot contain the content.
- **Hydration markers** flip `is_spa` only when the visible body text is also sparse (<400 chars), i.e. when the framework is present *and* the content is actually missing.

Found during a live audit of an Astro/Vercel site that was flagged as an SPA despite fully static output. The false positive skews audit findings (JS-rendering dependency is a scored factor) and slows every fetch. Builder-fingerprint logic (Wix/Webflow/Squarespace) is unchanged; genuinely client-rendered pages still render headlessly.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [x] Tested with a real URL before submitting (Astro/Vercel production site, static output correctly reports `is_spa: false`; empty-shell and sparse-hydration pages still report `true`)
- [x] SKILL.md files stay under 500 lines (if modified) — none modified
- [x] Python scripts output JSON (if modified) — `render_page.py` JSON output unchanged
- [x] Reference files stay under 200 lines (if modified) — none modified
- [x] `set -euo pipefail` used in any new shell scripts — no new shell scripts

`tests/test_render_page.py` gains coverage for both tiers; full file passes (43 tests).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- `render_page.py` no longer misclassifies SSR/SSG framework pages (Next.js, Nuxt, Svelte,
+  Astro islands) as SPAs. Hydration markers now flip `is_spa` only when the visible body
+  text is also sparse; genuinely empty mount shells and noscript walls still trigger
+  headless rendering unconditionally.
+
 ## [2.2.4] - 2026-07-20
 
 Community maintenance release following a full review of every open issue and pull request.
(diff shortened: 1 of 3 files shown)


AgriciDaniel:
Thank you for the diagnosis and tests. The SSR/SSG hydration-marker fix was integrated with the broader renderer work in #261 and released in v2.2.5, including coverage for complete framework pages and sparse hydration shells. Closing this PR as superseded.

PR #259, opened 2026-08-25
fix(geo): stop conflating training crawlers with search crawlers

0101-Brain-Code:
## Summary

Two pairs of AI crawlers were being used interchangeably in the `seo-geo` skill, which
makes the audit assert things its evidence does not support.

**`GPTBot` was listed as "ChatGPT web search".** It is OpenAI's model-training crawler.
`OAI-SearchBot` is the crawler that determines ChatGPT Search citability. The table
implied a site blocking `GPTBot` was not citable in ChatGPT Search, when the opposite is
true — blocking `GPTBot` while allowing `OAI-SearchBot` leaves the site fully citable.

**`Google-Extended` was scoreable as a Google Search signal.** It governs Gemini and
Vertex AI training and grounding only. It does not affect inclusion in Google Search, AI
Overviews, or AI Mode, all of which are served from the `Googlebot` index. The existing
`seo-technical` "Key distinctions" section already stated this correctly, so the two
skills disagreed with each other.

### What changed

- Corrected both rows in the `seo-geo` crawler table.
- Added a claim-to-bot mapping table: for each claim an audit might make, which
  user-agent supports it and which does not.
- Added the missing `OAI-SearchBot` row to the `seo-technical` crawler table — that
  skill performs the robots.txt check but had no way to substantiate a citability claim.
- Corrected the "Allow for AI search visibility" recommendation in both the skill and
  `agents/seo-geo.md`; `GPTBot` and `CCBot` are a licensing decision, not a
  search-visibility one.
- Output section now requires training access and search citability to be reported as
  separate findings rather than merged into one line.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [x] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Tested with a real URL before submitting — **N/A, see note below**
- [x] SKILL.md files stay under 500 lines (if modified) — `seo-geo` 344, `seo-technical` 228
- [ ] Python scripts output JSON (if modified) — no Python changed
- [ ] Reference files stay under 200 lines (if modified) — no reference files changed
- [ ] `set -euo pipefail` used in any new shell scripts — no shell scripts added

**On the real-URL checkbox:** this PR changes skill instructions and reference tables
only — there is no code path a URL would exercise. What I did run instead: the full
suite (398 passed, 1 skipped) plus 5 new tests in `tests/test_ai_crawler_claims.py`
pinning each corrected claim. The two pre-existing `tests/test_sync_flow.py` failures in
my environment are network/TLS related and unrelated to this change; they are deselected
in that count.

The factual claims are checkable against OpenAI's and Google's own crawler documentation
rather than against a crawl.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- AI crawler claims are now checked against the crawler that actually governs them.
+  `GPTBot` was documented as "ChatGPT web search" in the `seo-geo` crawler table; it is
+  OpenAI's model-training crawler, while `OAI-SearchBot` is what determines ChatGPT
+  Search citability. `Google-Extended` governs Gemini/Vertex training and grounding only
+  and is no longer treated as a Google Search readiness signal (Google Search, AI
+  Overviews, and AI Mode all follow `Googlebot`). Adds a claim-to-bot mapping table, the
+  missing `OAI-SearchBot` row to the `seo-technical` crawler table, and requires training
+  access and search citability to be reported as separate findings.
+
 ## [2.2.4] - 2026-07-20
 
 Community maintenance release following a full review of every open issue and pull request.
--- a/agents/seo-geo.md
+++ b/agents/seo-geo.md
@@ -27,7 +27,12 @@ You are a Generative Engine Optimization (GEO) specialist. When given a URL:
 
 ## AI Crawlers to Check in robots.txt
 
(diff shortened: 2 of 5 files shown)


AgriciDaniel:
The corrections are right; the review confirmed them against OpenAI's bots page (GPTBot trains, OAI-SearchBot decides ChatGPT search inclusion) and Google's crawler overview (Google-Extended does not affect Search inclusion). Three asks: cite those two vendor pages in the reference file, add Claude-SearchBot and Applebot-Extended, which are absent from the repo entirely, and fix the ClaudeBot line in seo-technical that still contradicts seo-geo. The tests assert exact wrapped strings including embedded newlines, which will break on any reflow; matching on the sentence would be more robust.

AgriciDaniel:
Landed in v2.3.0 as commit 808d3cc and 078dfd2; vendor citations, Claude-SearchBot and Applebot-Extended rows, and the ClaudeBot correction were added. Thank you.

PR #260, opened 2026-08-25
fix(backlinks): enforce the Common-Crawl-only no-score rule

0101-Brain-Code:
## Summary

`skills/seo-backlinks/SKILL.md` already said not to produce a numeric health score when
only Common Crawl data is available. The rule was violated in a real audit anyway, and
the reason turns out not to be that it was skimmed past.

**The repo contradicted itself.** `skills/seo/references/free-backlink-sources.md:33`
said:

> When only Common Crawl is available, cap the maximum health score at 70/100 and note
> "limited to domain-level metrics" in the report.

That instructs a number — with a specific cap and a specific caption — in exactly the
case the skill forbids one. Faced with a prohibition and a competing actionable
instruction, the agent followed the actionable one. The output reads as "this backlink
profile is weak" when the truth is "we have no data."

### What changed

**1. The two files now agree.** `free-backlink-sources.md` no longer instructs a capped
score. Common Crawl supplies domain-level rank and presence signals and none of the
referring-domain quality, anchor, or toxicity data a health score claims to summarise,
so the guidance is now `Not Assessed` with the rank/presence data surfaced as
low-confidence findings.

**2. The skill states it unambiguously.** The rule is a bolded **MUST NOT** rather than
descriptive prose at the end of a section.

**3. It is enforced rather than trusted.** `scripts/validate_backlink_report.py` gains
`validate_source_score_consistency()`, which fails a report (`status: FAIL`) when:

- a numeric score is present but no scoreable source (Moz, Bing, or DataForSEO) supplied
  data — the Common-Crawl-only case; or
- a finding marked `source: not-assessed` carries a numeric `score`, `value`, or
  `health_score`.

String-encoded numbers (`"42"`) and errored-out sources (a Moz call that returned 401)
are both handled, and the check runs even when the caller supplies no `scoring_factors`
block at all. The skill now requires a passing validator run before any backlink output
is presented.

Worth noting: this validator already existed with a related data-sufficiency gate, but
nothing in the repo invoked it. That is the other half of why the instruction never held.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [x] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [ ] Tested with a real URL before submitting — **see note below**
- [x] SKILL.md files stay under 500 lines (if modified) — `seo-backlinks` 298
- [x] Python scripts output JSON (if modified) — `validate_backlink_report.py` unchanged in that respect
- [x] Reference files stay under 200 lines (if modified) — `free-backlink-sources.md` 119
- [ ] `set -euo pipefail` used in any new shell scripts — no shell scripts added

**On the real-URL checkbox:** the validator operates on a report JSON, not a URL. What I
ran instead, which I think is the more relevant check — a simulated credential-less run
against an empty `HOME`:

```
$ backlinks_auth.py --check --json   # moz: unavailable, bing: unavailable, commoncrawl: available
$ validate_backlink_report.py --report cc-only-with-score.json --json
status: FAIL
  health_score       :: Numeric score (58) produced when Common Crawl was the only source available…
  findings[0].score  :: Finding 'Backlink Health' is source: not-assessed but carries a numeric score (58)…
$ validate_backlink_report.py --report cc-only-not-assessed.json --json
status: PASS
```

Plus the full suite (403 passed, 1 skipped) and 10 new cases in
`tests/test_backlink_score_gate.py`. The two guards against the reference-file
contradiction were verified by restoring the original line and confirming they fail,
rather than trusting a green run. The two pre-existing `tests/test_sync_flow.py` failures
in my environment are network/TLS related and unrelated to this change; they are
deselected in that count.

## Note for the maintainer

This touches `CHANGELOG.md` under a new `## [Unreleased]` heading, as does my other open
PR (#259). Whichever merges second will conflict there — happy to rebase that one, or
drop the changelog hunk from either if you would rather batch them yourself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- The Common-Crawl-only no-score rule is now enforced instead of merely stated. The
+  `seo-backlinks` skill said not to produce a numeric health score when only Common Crawl
+  data is available, while `skills/seo/references/free-backlink-sources.md` told the agent
+  to "cap the maximum health score at 70/100" in the same case. The reference gave the more
+  actionable of two contradictory instructions, which is the likely reason the rule was
+  violated in practice. The guidance now agrees across both files, and
+  `validate_backlink_report.py` gains a `source_score_consistency` check that fails a
+  report carrying a number with no scoreable source (Moz, Bing, or DataForSEO), or a
+  `source: not-assessed` finding carrying a score.
+
 ## [2.2.4] - 2026-07-20
 
 Community maintenance release following a full review of every open issue and pull request.
--- a/scripts/validate_backlink_report.py
+++ b/scripts/validate_backlink_report.py
@@ -235,6 +235,102 @@ def validate_health_score(scoring_factors: dict) -> list:
     return issues
 
 
+# Sources that can support a numeric backlink health score. Common Crawl is
+# deliberately absent: it provides rank/presence signals only, not the
+# referring-domain quality, anchor, or toxicity data a score implies.
+SCOREABLE_SOURCES = ("moz_data", "bing_data", "dataforseo_data")
+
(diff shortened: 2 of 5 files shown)


AgriciDaniel:
Landed in v2.2.6 as commit 37bd5d and 54f0e9 and 50b382 (cherry-picked with -x, your authorship preserved). Thank you.

PR #262, opened 2026-08-25
Add Keywords Everywhere (Open PageRank) as a backlinks fallback source

MediaJohnD:
## Summary
- Adds Keywords Everywhere (formerly OpenPageRank) as an optional, free-signup backlink data source in `seo-backlinks`, following the existing `moz_api.py`/`backlinks_auth.py` credential and source patterns
- Provides a single 0-10 domain rank metric as a lightweight Profile Overview fallback when Moz isn't configured (confidence 0.60, below Moz's 0.85, above Common Crawl's 0.50) — no referring domains/anchors/top pages, use Moz for those
- New `scripts/keywordseverywhere_api.py` client normalizes and SSRF-validates every input domain via the existing `url_safety` module before querying

## Notes
- Reviewed over two rounds by an independent agent review; caught and fixed a scheme-validation gap (`ftp://localhost`-style bypass) and a silent >100-domain truncation before landing
- SKILL.md version left at the repo's locked 2.2.4 to match `plugin.json` and every other skill

## Test plan
- [x] `python -c "import ast; ast.parse(...)"` syntax check on both edited/new Python files
- [ ] Live API call against a real Keywords Everywhere key (not run here — no test key available in this environment)

### DIFF
--- a/scripts/backlinks_auth.py
+++ b/scripts/backlinks_auth.py
@@ -44,6 +44,7 @@ CACHE_DIR = os.path.expanduser("~/.cache/claude-seo/commoncrawl")
 SERVICE_AUTH = {
     "moz": "api_key",
     "bing": "api_key",
+    "keywordseverywhere": "api_key",
     "commoncrawl": "none",
     "verify": "none",
 }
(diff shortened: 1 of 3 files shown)


AgriciDaniel:
AgriciDaniel/claude-seo

PR #263, opened 2026-08-26
fix(content-quality): tokenise CJK so Korean/Japanese/Chinese pages get real scores

ufo6849-web:
## The problem

`content_quality.py` tokenises with `_TOKEN_RE = re.compile(r"[A-Za-z][A-Za-z'\-]*")`.

Hangul, kana and ideographs match nothing, so **CJK text produces zero tokens**. I measured the same sentence in both languages against v2.2.5:

| input | tokens | unique | density | overall |
|---|---|---|---|---|
| English | 58 | 53 | 0.172 | 70 |
| Korean (same content) | **0** | **0** | **0.0** | 65 |

With `tokens == 0`, `information_density` collapses to `0.0`, `_repetition_score` returns early on `len(tokens) < 4`, and the phrase lists are English — so **every Korean page lands on an identical `overall_quality: 65` with `low-density` + `thin-content`**, no matter what it says.

Content Quality carries the **largest weight in the SEO Health Score (23%)**, so that axis is currently inert for any non-Latin site. `/seo content` and `/seo audit` cannot distinguish a good Korean page from a bad one.

## The change

Three alternations added to the tokeniser:

- **Hangul syllable runs** — Korean is space-delimited, so a run of syllables is one word.
- **kana** and **CJK ideographs** — Chinese and Japanese are unspaced, so one character counts as one token.

Latin tokenisation is byte-for-byte untouched, and a regression test pins it to the original regex's output.

## Result

The same three Korean pages now differentiate instead of collapsing:

| page | before | after |
|---|---|---|
| `/about` | 65 (0 tokens) | **68** (300 tokens) |
| `/learn/…` | 65 (0 tokens) | **70** (273 tokens) |
| `/today` | 65 (0 tokens) | **72** (443 tokens) |

`thin-content` now actually discriminates: it clears on two of the three and stays on the shortest.

## Known limit, documented in the diff

`_ENTITY_RE` is a capitalised-name heuristic, and CJK scripts have no letter case — so entity detection stays at 0 for these languages and density still skews low. I left a `NOTE` beside it rather than papering over it: tokenisation is the load-bearing fix, and CJK named-entity detection needs a model, not a regex.

## Tests

4 added to `tests/test_content_quality.py` following the existing style — Korean tokenisation, JA/ZH tokenisation, document discrimination, and a Latin regression guard.

- Full file: **31 passed**.
- Mutation-checked: reverting the tokeniser fails exactly the 3 CJK tests, and the Latin guard passes both ways as intended.

Tested against a live Korean site (`kongmingsaju.com`) via `render_page.py` → `content_quality.py`, per CONTRIBUTING.

### DIFF
--- a/scripts/content_quality.py
+++ b/scripts/content_quality.py
@@ -141,11 +141,23 @@ _AI_PATTERNS: tuple[str, ...] = (
 )
 
 
-_TOKEN_RE = re.compile(r"[A-Za-z][A-Za-z'\-]*")
+# Latin words, Hangul eojeol, and CJK/kana characters. Korean is space
+# delimited so a run of syllables is one token; Chinese and Japanese are
+# unspaced, so one ideograph or kana counts as one token.
+_TOKEN_RE = re.compile(
+    r"[A-Za-z][A-Za-z'\-]*"
+    r"|[\uac00-\ud7a3]+"
+    r"|[\u3041-\u3096\u30a1-\u30fa\u30fc]"
+    r"|[\u3400-\u4dbf\u4e00-\u9fff]"
+)
 _NUMBER_RE = re.compile(r"\b\d+(?:[.,]\d+)?(?:%|st|nd|rd|th)?\b")
 # Capitalised multi-word names: rough proper-noun heuristic. Two or more
 # capitalised tokens in a row count as one entity.
 _ENTITY_RE = re.compile(r"\b(?:[A-Z][a-z]+(?:\s+[A-Z][a-z]+)+)\b")
+# NOTE: CJK scripts have no letter case, so this heuristic finds no entities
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
English output is byte-identical at both heads and the CJK collapse is genuinely fixed. Two follow-ups before the score is comparable: entity density stays 0.0 for CJK (the entity regex is case-based and the word boundary never fires next to CJK), and the English-only phrase lists give CJK pages a free 100 on half the composite weight. The smallest honest change is a coverage field in the output that says which signals were computed for the detected script, so a CJK score is not presented as comparable to an English one.

AgriciDaniel:
Landed in v2.3.0 as commit 60a532a; a coverage object marks which signals were not computed for CJK text. Thank you.

PR #269, opened 2026-08-27
Improvement to the seo-hreflang skill for detecting tags in the head section

tetus129-bit:
## Summary

The current agent is restrictive when searching for hreflang tags within the head section. It searches for them in exactly this order:

`<link rel="alternate" hreflang="en-US" href="https://example.com/page" />`

But actually, the order doesn't matter; it could also be:

`<link href="https://example.com/page" hreflang="en-US" rel="alternate">`

This update prevents false positives when detecting hreflang tags in the head section. With the original version, if the order of the tags is different... the tool fails to detect them and flags them as a critical error, even though the hreflang tags are actually present.

## Type of Change

- [x ] Bug fix

## Checklist

- [x ] Tested with a real URL before submitting

### DIFF
--- a/skills/seo-hreflang/SKILL.md
+++ b/skills/seo-hreflang/SKILL.md
@@ -117,6 +117,10 @@ Best for: Sites with <50 language/region variants per page.
 
 Place in `<head>` section. Every page must include all alternates including itself.
 
+The order doesn't matter; it can also be:
+
+<link href="https://example.com/page" hreflang="en-US" rel="alternate">
+
 ### Method 2: HTTP Headers
 Best for: Non-HTML files (PDFs, documents).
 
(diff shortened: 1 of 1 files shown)


AgriciDaniel:
The described defect is not reproducible: scripts/parse_html.py looks attributes up by name, so order and self-closing forms parse (13 of 14 variants tested). What does fail is case: rel="Alternate" and REL="Canonical" are dropped because rel tokens are matched case-sensitively. If you retarget this PR at parse_html.py (lower-case the rel tokens before comparing) with a test for the mixed-case form, it becomes a real fix; the six doc lines on their own do not change behaviour.

AgriciDaniel:
Closing with thanks: the investigation your report prompted found the real defect, case-sensitive rel token matching in parse_html.py, which v2.3.0 fixes (rel="Alternate" and REL="Canonical" are now detected). The changelog credits this PR.

PR #275, opened 2026-08-31
feat(matomo): Matomo Reporting API extension (GA4 alternative or complement)

tim-peters:
## Summary

Adds an optional Matomo extension providing the Matomo Reporting API as a GA4 alternative or complement for self-hosted and Matomo Cloud instances: `/seo matomo organic | top-pages | device | country | referrers | keywords`, a live credential probe, and a conditional `seo-matomo` audit agent that writes `findings/matomo.md` during `/seo audit` (alongside `seo-google` when both GA4 and Matomo are configured).

## What's included

- **`scripts/matomo_auth.py`** — credential management (`~/.config/claude-seo/matomo.json` + `MATOMO_URL` / `MATOMO_API_TOKEN` / `MATOMO_SITE_ID` env fallbacks), live `--check` probe via `API.getMatomoVersion`, `--setup` instructions
- **`scripts/matomo_report.py`** — Reporting API client: organic trend (`segment=referrerType==search`), entry-page landing pages, device / country breakdowns, referrer channels + search engines, keywords with anonymized-share flagging. POST-only auth (token never in URLs or logs), standard `{status, data, error, metadata}` envelope, `--json` output, both registered in the runtime allowlist
- **`extensions/matomo/`** — `install.sh` / `install.ps1` / `uninstall.sh` (atomic 0600 settings writes), skill, setup docs, audit agent
- **Audit integration** — conditional spawn in `skills/seo` + `skills/seo-audit` when `matomo_auth.py --check` succeeds
- **Docs** — README (commands + extensions + count 8→9), AGENTS.md, CLAUDE.md, `docs/COMMANDS.md`, PRIVACY.md (data-transmission rows incl. the relaxed-URL rationale), CHANGELOG (Unreleased)
- **Tests** — 29 unit tests in `tests/test_matomo_report.py` (envelope shape, token redaction, credential isolation, all report parsers); `("matomo", "seo-matomo")` added to the extension-structure tests

## Design notes

- **Self-hosted on private networks / localhost is intentionally supported**: the instance URL gets a light sanity check (http/https scheme + host, no userinfo) instead of the strict `url_safety.validate_url()` SSRF protection used for arbitrary web fetches. Rationale is documented in the script docstring and PRIVACY.md — self-hosted Matomo frequently lives on internal networks, where strict validation would block legitimate setups.
- **Form-tolerant parsing** (verified live against Matomo 5.13): multi-row DataTables arrive as JSON arrays, date summaries as objects; rates are computed from raw counts (`bounce_count` / `entry_bounce_count`) instead of parsing Matomo's locale-dependent formatted metric strings (`"7\u00a0%"`).
- **Correct Matomo 4/5 method name**: `Referrers.getReferrerType` (singular — `getReferrersType` does not exist); machine-readable channel codes come from the row `segment` field since labels are localized ("Suchmaschinen", "Direkte Zugriffe", …).
- **Anonymized keywords** are detected locale-independently via the row `segment` field (`referrerKeyword==` with empty value) plus known label variants ("Keyword not defined", "Suchbegriff nicht definiert", …), and collapse into a single `(not provided)` row with the anonymized share surfaced.

## Testing

- `pytest tests/` (targeted: manifest consistency, extension structure, matomo report) — 190 passed; test isolation ensures no real credentials/network are touched
- `py_compile` over all scripts; `portability_check.py` clean (0 findings, 34 SKILL.md files)
- Live-verified against a self-hosted Matomo 5.13 instance: all 7 subcommands, plus negative cases (invalid token → auth error without token leak, invalid idSite → verbatim Matomo error, unreachable URL → connection-error class)
- `bash -n` on installers, executable bits set, installer follows the safe quoted-heredoc + atomic 0600 write pattern


AgriciDaniel:
Thank you, Tim. All four of your commits are on the v2.4.0 release line with your authorship kept, and the v2.4.0 CHANGELOG credits you for #275. We added a few follow-up fixes on top (url_safety routing for every instance request, the credential file in ~/.config/claude-seo, the agent contract). GitHub showed this PR as conflicting only because the cherry-picked commits have new SHAs. Shipped in v2.4.0, so closing.

PR #279, opened 2026-09-01
feat: strip invisible Unicode watermarks in content_humanize

support224:
## Summary

`content_humanize.py` removed AI-typical phrasing but was blind to invisible Unicode watermarks — zero-width characters, directional marks/overrides, and Unicode tag characters that some AI tools embed in generated text. Worse, zero-width codepoints inside words silently defeat the existing `\b`-boundary replacements, so watermarked text also escaped the phrasing pass.

This PR adds a `strip_invisible()` pass that runs before the phrasing replacements:

- Removes zero-width characters (ZWSP, ZWNJ, WJ, BOM), directional marks/overrides, and Unicode tag characters
- Normalizes exotic spaces (NBSP, thin/hair spaces, etc.) to regular spaces
- **Preserves emoji sequences** — ZWJ and variation selectors adjacent to emoji are kept so 👨‍👩‍👧 and ❤️ survive intact
- Reports what was stripped in the JSON output (counts per category)

The `seo-content` SKILL.md gains routing triggers (`humanize`, `remove watermarks`, `invisible characters`) and documents the scope limits honestly: statistical/probability-distribution watermarks are out of reach for character stripping, and the tool is intended for cleaning the user's own drafts.

## Type of Change

- [ ] Bug fix
- [x] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [x] Tested with a real URL before submitting (script tested on real watermarked samples; 9 new unit tests pass)
- [x] `set -euo pipefail` used in any new shell scripts (n/a — no shell scripts)

## Notes for reviewers

- The emoji-preservation logic is the subtle part: ZWJ and variation selectors are only kept when adjacent to an emoji-range codepoint — a bare ZWJ between Latin letters (the watermark case) is stripped. `tests/test_content_humanize.py` covers both directions.
- Stripping runs first so the downstream phrasing regexes see clean word boundaries; test `test_zero_width_inside_word_defeats_boundaries` locks in the ordering.


AgriciDaniel:
Stripping zero-width watermark characters is a good addition. The adjacency check treats every code point at or above U+2100 as emoji, which is above Arabic and Devanagari, so Persian ZWNJ (می‌رود becomes میرود) and Devanagari ZWNJ and ZWJ are deleted. That is silent text corruption in a tool that promises not to alter meaning. Please use unicodedata categories or an explicit emoji-range test for the adjacency check, keep ZWNJ and ZWJ when a neighbour is a letter in a script that uses them, and add those two cases to the tests.

AgriciDaniel:
Landed in v2.3.0 as commit 3a13b30 and 2d34cc1; ZWNJ and ZWJ are preserved next to letters in scripts that use them. Thank you.

PR #284, opened 2026-09-04
docs(seo-image-gen): document an optional HTTP-only image provider for MCP-less environments

binyangzhu000-sudo:
## Summary

`skills/seo-image-gen` depends entirely on the banana / Gemini MCP server. The
`MCP unavailable` row in its error table currently has no resolution for
environments where an MCP server cannot be installed at all — CI runners,
containers, locked-down machines — it just points back at the installer.

This PR documents an **optional** HTTP-only fallback for exactly that case and
changes no defaults.

- `skills/seo-image-gen/references/atlas-cloud.md` (new, 146 lines, under the
  200-line reference limit): submit → poll → download recipe, the SEO use case →
  `size` mapping mirrored from `SKILL.md`'s use-case table, model list with
  reference prices, and an error table.
- `skills/seo-image-gen/SKILL.md` (+6/−1): the `MCP unavailable` error row, a
  two-line note under Prerequisites, and the reference index. Nothing else moves;
  banana / Gemini stays the recommended path everywhere.

Repo conventions followed:

- Keys read from the environment only, `<atlascloud-api-key>` placeholder in the
  docs — no key material in the diff.
- The download step explicitly routes through `scripts/url_safety.py`
  (`validate_url()` / `safe_requests_session()`) per CONTRIBUTING rule 3, even
  though the URL comes back from the API rather than from the user.
- `SKILL.md` stays well under 500 lines (178); the new reference is cited from
  `SKILL.md`, so the orphan-reference guard in `tests/test_manifest_consistency.py`
  is satisfied.
- No script is vendored, so the "claude-seo does not vendor a local generation
  fallback script" position is preserved — this is documentation of an external
  API, not a bundled generator.

`extensions/banana/skills/seo-image-gen/SKILL.md` (the installed-extension mirror)
is deliberately left untouched: it ships without a `references/` directory, so the
pointer would not resolve there. Happy to add an inlined variant if you'd prefer
the mirror to carry it too.

## Validation

- Live end-to-end check of the documented flow: `POST /api/v1/model/generateImage`
  with `alibaba/wan-2.7/text-to-image` → prediction id → bounded polling →
  `status: completed` → PNG output URL. The `size` gotcha documented in the
  reference (`1024*1024`, not `1024x1024`) is a real 400 I hit while verifying.
- `git diff --check` clean; secret scan over the staged diff clean.
- Markdown-only change; no Python touched, so the CI syntax check and pytest
  suite are unaffected.

---

### 🤝 Partnership & contact

This PR comes from the **Atlas Cloud** team. Beyond the integration above, we'd love to explore a closer collaboration with **claude-seo** — for example **co-marketing** or a featured integration.

If that sounds interesting, reach out anytime:
- 📧 **marketing@atlascloud.ai**
- 🌐 https://www.atlascloud.ai
- 📚 Docs: https://www.atlascloud.ai/docs

And of course, happy to revise this PR to match your project's conventions — just leave a comment. 🙌

### DIFF
--- a/skills/seo-image-gen/SKILL.md
+++ b/skills/seo-image-gen/SKILL.md
@@ -38,6 +38,10 @@ is connected by checking if `gemini_generate_image` or `set_aspect_ratio` tools 
 available. If tools are not available, inform the user the extension is not installed
 and provide install instructions.
 
+Environments where no MCP server can be installed (CI runners, containers) have an
+optional HTTP-only path documented in `references/atlas-cloud.md`. It is a fallback,
+not a default — prefer the banana MCP whenever it is available.
+
 ## Quick Reference
 
 | Command | What it does |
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
Declining: this places vendor documentation inside a core skill on documentation alone, with a bare curl recipe and a UTM-tagged link. Reopen condition: an extensions/atlascloud tree with a client that goes through url_safety, tests that run without a key, a PRIVACY.md row, and no tracking parameters.

From the 2026-09-10 full review and the Phase 3 decisions; the conditions above are what would reopen this.

PR #285, opened 2026-09-06
fix(hooks): detect every ld+json block and stop flagging template expressions

alebgl77:
## Summary

`hooks/validate-schema.py` only matched `<script type="application/ld+json">` when `type` was the first and only attribute, and it reported runtime template expressions in framework files as invalid JSON. This PR fixes both and widens the accepted `@context` forms.

Reproduced on `main` (a1480c7) by running the hook directly on fixture files.

**Silent skips (exit 0, nothing validated):**

| Fixture | Result on main |
|---|---|
| `<script type="application/ld+json" nonce="abc">` containing `[Business Name]` | exit 0, placeholder not caught |
| `<script id="ld" type="application/ld+json">` containing `ClaimReview` | exit 0, retired type not caught |
| `<script type=application/ld+json>` (unquoted) | exit 0 |

CSP nonces are standard on production sites, so the hook was skipping exactly the blocks it exists to check.

**Noise on every edit (exit 1 warning):**

| File | Body | Result on main |
|---|---|---|
| `.tsx` | `{JSON.stringify(schema)}` | Invalid JSON |
| `.vue` | `{{ jsonLd }}` | Invalid JSON |
| `.php` | `<?php echo json_encode($schema); ?>` | Invalid JSON |
| `.ejs` | `<%- JSON.stringify(schema) %>` | Invalid JSON |
| `.svelte` | `{@html ...}` | Invalid JSON |

The hook's own extension list targets these frameworks, and nothing the author does can make a runtime expression parse as JSON.

**Valid `@context` reported as wrong:** `"https://schema.org/"` (trailing slash), `["https://schema.org", {...}]`, `{"@vocab": "https://schema.org/"}`.

**Changes**

- Match every `<script ...>...</script>` pair and decide on the `type` attribute alone, so attribute order, extra attributes, tag case and unquoted values no longer matter. Other script types are still ignored.
- Skip server template markers (`<?php`, `<?=`, `<%`) in every file type. Skip component expressions (JSX, Vue, Svelte, template literals) only in `.jsx/.tsx/.vue/.svelte`, so a malformed object literal in a plain `.html` file is still reported.
- Accept the schema.org context as a string with or without trailing slash, as a list containing it, or as an object whose `@vocab` is it. Foreign contexts are still reported.

`validate_jsonld()` gains an optional `filepath` argument; `main()` passes the resolved path. No callers outside the hook.

**Tests**

`tests/test_schema_hook_detection.py`, 18 cases: 6 false negatives, 8 template cases including the two guards (literal JSON in a `.tsx` still validated, malformed literal in `.html` still reported), 4 `@context` forms.

```
pytest tests/test_schema_hook_detection.py tests/test_schema_hook_policy.py tests/test_cross_platform_hooks.py tests/test_schema_v2.py
51 passed
ruff check hooks/validate-schema.py tests/test_schema_hook_detection.py
All checks passed
```

Note for reviewers: the full suite does not collect on a machine without the Google client libraries because `scripts/gsc_query.py` calls `sys.exit(1)` at import and `tests/test_gsc_pagination.py` imports it at module level. Unrelated to this change; happy to open a separate PR that guards the import if useful.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

- [x] Tested with a real URL before submitting (hook exercised on fixture files covering every case above; the hook does not fetch URLs)
- [x] SKILL.md files stay under 500 lines (if modified) - not modified
- [x] Python scripts output JSON (if modified) - hook output format unchanged
- [x] Reference files stay under 200 lines (if modified) - not modified
- [x] `set -euo pipefail` used in any new shell scripts - no shell scripts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- The JSON-LD hook now validates every `application/ld+json` block regardless of
+  attribute order, CSP `nonce`, `id` or `data-*` attributes, tag case, or an
+  unquoted type value; such blocks were previously skipped without validation.
+- The JSON-LD hook no longer reports runtime template expressions (JSX, Vue,
+  Svelte, template literals in component files; PHP and EJS everywhere) as
+  invalid JSON, while a malformed literal in plain HTML is still reported.
+- The JSON-LD hook accepts the schema.org `@context` with a trailing slash, in
+  list form, and in `{"@vocab": ...}` object form.
+
 ## [2.2.5] - 2026-08-25
 
 ### Added
--- a/hooks/validate-schema.py
+++ b/hooks/validate-schema.py
@@ -49,6 +49,45 @@ BRACKET_PLACEHOLDERS = (
 )
 BARE_PLACEHOLDER_RE = re.compile(r"\bREPLACE(?:_[A-Z]+)*\b")
 
+# Match every <script ...>...</script> pair, then filter on the type attribute.
+# The previous pattern required ``type`` to be the first and only attribute, so
+# blocks carrying a CSP ``nonce``, an ``id`` or ``data-*`` attributes, or an
+# unquoted type value were skipped without validation.
+SCRIPT_TAG_RE = re.compile(
+    r"<script\b([^>]*)>(.*?)</script\s*>", re.DOTALL | re.IGNORECASE
+)
+LD_JSON_TYPE_RE = re.compile(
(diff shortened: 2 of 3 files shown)


AgriciDaniel:
The new coverage is real and wanted (nonce and id blocks now validated, JSX {JSON.stringify(schema)} no longer flagged). One regression: SCRIPT_TAG_RE uses [^>]* for attributes, so any attribute value containing > (for example data-cond="a>b" or a templated CSP nonce) truncates the tag and a valid block reports Invalid JSON. A legitimate headline containing [INSERT] behind a nonce goes from exit 0 to exit 2, which blocks an edit main allows. A small tokenizer that respects quoted attribute values fixes it; the fixtures from the review are available on request.

AgriciDaniel:
Landed in v2.3.0 as commit 9aefba1 and 697b92f and da0fd00; the attribute tokenizer now respects quoted values and no longer backtracks on unclosed tags. Thank you.

PR #287, opened 2026-09-06
fix(content_humanize): four replacements that produce broken English

Ilno:
## Summary

`content_humanize` rewrites prose that ships to a live site, so a wrong replacement is not a crash: it is broken English published to readers. The module has no test file, which is why these went unnoticed.

All four reproduce on current `main`:

| Input | `main` today | This PR |
|---|---|---|
| `Leverage the power of AI to unlock the potential of your CRM.` | `Use AI to use of your CRM.` | `Use AI to get more from your CRM.` |
| `In conclusion, this is fine.` | `this is fine.` | `This is fine.` |
| `This is a game-changer.` | `this is a important.` | `This is a big change.` |
| `First and foremost, delve into the data.` | `First,explore the data.` | `First, explore the data.` |

## The four causes

**1. `unlock the potential of X` leaves the `of` behind.** The pattern consumes `potential` but not the preposition, so `use` lands in front of an orphan `of`. Split into an `-of` variant (`get more from`) and a bare one (`do more`).

**2. A deleted sentence opener takes the capital with it.** Every deletion-style entry (`in conclusion,`, `moreover,`, `ultimately,`, `in essence,` …) has this at a sentence start. Added a recapitalisation pass that runs *only* when something was actually deleted, and that masks fenced blocks, inline code and URLs first, since an uppercase letter changes their meaning there.

**3. `game-changer` is a noun, `important` is an adjective.** No article logic can rescue `a game-changer -> a important`, because the part of speech changed. The replacement is now a noun (`big change`, plural `big changes`). Separately added an `a`/`an` agreement helper for the cases where a replacement legitimately changes the following initial, e.g. `an ever-evolving field -> a changing field`.

**4. Two patterns eat the trailing space without putting one back.** `first and foremost,?\s*` consumes the whitespace but replaces with `"first,"`. Same for `last but not least`. Thousands separators (`1,000`) are unaffected.

## Tests

Adds `tests/test_content_humanize.py`, 19 tests: the four regressions above, plus the safety properties the fixes depend on — URLs and fenced code are not recapitalised, unknown idiom is left alone, no capitalisation pass runs when nothing was deleted, and the change log still counts every replacement.

Tested per CONTRIBUTING (step 4) on real drafts in both directions before and after.

## Verification

```
pytest tests/ -q
460 passed in 113.02s
```

441 before this PR, 460 after (+19), no regressions.


Ilno:
Closing this one, sorry for the noise. Opened it without the go-ahead on my side; the analysis is in the thread if it is useful to you, feel free to take any of it.

PR #288, opened 2026-09-06
fix(agents): open the findings file first so a turn cap loses partial work, not all of it

### LINKED ISSUE #177: Full audit: subagents write no findings files on large sites (low maxTurns + conditional/missing persistence)
## Summary

On a large programmatic site (~1,755 URLs), running `/seo audit` produced **zero findings files**. All 8 spawned specialist subagents completed their research but wrote nothing to `output_dir/findings/`. The final report had no per-category inputs.

After inspecting the installed agent definitions in `~/.claude/agents/`, this traces to two independent causes in the agent frontmatter/instructions, both of which get worse as site size grows.

Tested on **v2.2.4**, manual install (`install.sh`), macOS, Claude Code with subagent delegation enabled.

## Cause 1 — `maxTurns` is too low for site-scale audits

Every agent hit its turn ceiling mid-research, before reaching its write step. Because the write instruction is the *last* thing in each agent, the cap is reached during fetching/analysis and the agent terminates with `completed` status having written nothing.

Observed tool-call counts vs. the configured cap:

| Agent | `maxTurns` | tool calls before stop |
|---|---:|---:|
| seo-schema | 15 | 15 (hit exactly) |
| seo-content | 15 | 18 |
| seo-performance | 15 | 17, then 32 on resume |
| seo-sitemap | 15 | 20 |
| seo-technical | 20 | 21 |
| seo-local | 20 | 24 |
| seo-sxo | 20 | 32 |

The caps are reasonable for auditing a single page, but a full-site audit spends 15+ turns just fetching a representative URL sample (the sitemap here had 1,685 event pages across 3 templates) before any analysis begins. Resuming a stopped agent didn't help — it got a fresh budget, re-researched, and hit the same wall in the same place.

## Cause 2 — the persistence instruction is conditional, last, and often absent

For the agents that *do* mention writing, the instruction is gated and placed at the very end:

> `## Persistence Contract`
> `If output_dir is provided by the audit orchestrator, write:`

Framed this way, an agent that runs out of turns — or that reads the gate as optional — treats the file as an afterthought and returns a chat summary instead.

More significantly, **13 of 18 `seo-*` agents have no persistence instruction at all**, including `seo-geo`, `seo-local`, `seo-sxo`, and `seo-sitemap`. Those agents were never going to write a findings file regardless of the turn limit:

```
seo-backlinks  seo-cluster  seo-dataforseo  seo-drift  seo-ecommerce
seo-flow  seo-geo  seo-google  seo-image-gen  seo-local  seo-maps
seo-sitemap  seo-sxo
```

## Suggested fix

Three changes to the `seo-*` agent definitions:

1. **Raise `maxTurns`** to something realistic for site-scale work (~60), or scale it with crawl size.
2. **Make persistence unconditional and universal** — every audit agent should write its `findings/<category>.md`, not gate it behind an orchestrator convention, and no audit agent should be missing the instruction.
3. **Move the write to the front and make it incremental** — instruct each agent to create its findings file up front and append as it confirms each finding, so a turn-cap termination leaves partial results on disk instead of nothing. This makes the failure mode graceful rather than total.

Happy to send a PR along these lines if it's a welcome direction.

## Note

The Python tooling itself worked well throughout — `render_page.py --mode always` was essential for a site behind SiteGround's UA-based bot blocking (generic `python-requests` UA gets a 403; the Playwright path with a real Chrome UA gets 200), and sitemap discovery was accurate. This report is specifically about the subagent orchestration layer.

### LINKED ISSUE #272: Audit subagents hit maxTurns before writing their findings file, losing the work
## Summary

On a real `/seo audit` run, **10 of 11 spawned subagents hit their `maxTurns` cap before writing their findings file**. The analysis work was done — tool calls ran, data was fetched, some agents even saved raw JSON — but the deliverable file was never written, so from the orchestrator's perspective the work was simply lost.

Every agent had to be manually resumed with an explicit "stop investigating, write what you have now" instruction to recover anything.

## Environment

- claude-seo **v2.2.5** (marketplace plugin install)
- Claude Code 2.1.177, Windows 11
- `claude-seo doctor --json` → `ready: true`, `browser_ready: true`, `mode: plugin-fallback`, Python 3.14
- Google API Tier −1 (no credentials), backlinks Tier 0
- Target: a small Blogger site — **5 pages total** (home, 2 posts, 2 static pages)

Worth emphasizing: this was a *tiny* site. If 5 pages exhausts the turn budget, a real audit target will not fare better.

## Data

Numbers are at the moment each agent first stopped, before any resume.

| agent | `maxTurns` | tool uses | tokens | wrote findings file? |
|---|---:|---:|---:|---|
| `seo-cluster` | 20 | 66 | 248,101 | ✗ |
| `seo-technical` | 20 | 38 | 83,612 | ✗ |
| `seo-schema` | 15 | 25 | 77,511 | ✗ |
| `seo-sxo` | 20 | 38 | 76,081 | ✗ |
| `seo-performance` | 15 | 27 | 69,477 | ✗ |
| `seo-drift` | 15 | 20 | 66,926 | ✗ |
| `seo-sitemap` | 15 | 30 | 57,517 | ✗ |
| `seo-content` | 15 | 20 | 41,462 | ✗ |
| `seo-geo` | 20 | 24 | 39,875 | ✗ |
| `seo-visual` | 15 | 36 | 29,881 | ✗ |
| `seo-backlinks` | 20 | 29 | 62,557 | ✓ |

`seo-visual` is the clearest illustration: it captured and saved **all 8 screenshots** plus 6 raw JSON files, then ran out of turns before opening any of them. On resume it reported, accurately, that it had never looked at its own output.

## Root cause

Two things compound.

**1. `maxTurns` is below what the agent definition asks for.**

`agents/seo-technical.md` sets `maxTurns: 20` and instructs the agent to cover 9 categories — crawlability, indexability, security, URL structure, mobile, Core Web Vitals, structured data, JS rendering, IndexNow — then write a severity-tagged report. On resume it self-reported four categories as entirely uncovered.

**2. Turn budget is spent rediscovering the environment.**

Partial outputs captured at the stop point show agents opening with environment exploration rather than analysis:

- `seo-technical`: *"I'll start by exploring the audit environment, checking the available tooling, and setting up the output directory."*
- `seo-performance`: *"I'll start by verifying the environment, checking API credential status, and locating the claude-seo scripts."*
- `seo-sxo`: *"Found the skill. Let me read the reference files and check the CLI usage in parallel."*

Each agent independently rediscovers the scripts directory, `claude-seo run` usage, and credential tiers — facts the orchestrator already knows before spawning anything.

## Why this is worse than a slow audit

The failure mode is quiet. The orchestrator receives a "stopped at its N-turn limit (partial result)" note, but `findings/` is simply missing that file. Nothing says *which* categories went uncovered.

An orchestrator that doesn't check the filesystem will present an audit as complete when several categories were never analyzed, and the health score will be computed from whatever subset happened to survive. That is a correctness problem, not just a throughput one.

## Suggested fixes

**A. Write the findings file first, then refine it** — cheapest, and fixes the failure mode rather than the threshold.

Add to each `agents/*.md`: as the first action, `Write` a skeleton findings file with the section headers and `조사 중` / `in progress` placeholders; update it as the analysis proceeds. A turn-limit cut then leaves a partial-but-real report instead of nothing. This alone would have salvaged all 10 runs.

**B. Raise `maxTurns` on the heavy agents.** Current values: `seo-content`, `seo-schema`, `seo-sitemap`, `seo-performance`, `seo-visual`, `seo-drift`, `seo-google`, `seo-flow`, `seo-image-gen` at 15; `seo-technical`, `seo-geo`, `seo-sxo`, `seo-cluster`, `seo-backlinks`, `seo-local`, `seo-ecommerce` at 20; `seo-maps`, `seo-dataforseo` at 25. Based on the observed distribution, `seo-technical` and `seo-cluster` in particular look under-budgeted by roughly 2×.

**C. Have `skills/seo-audit` pass environment context into each subagent prompt** — the resolved scripts path, that `claude-seo run` is the entry point, the credential-tier output from `google_auth.py --check` and `backlinks_auth.py --check`, the discovered URL list, and Playwright readiness. Observationally this is worth 3–5 turns per agent.

**D. Instruct agents to mark unreached items explicitly** rather than filling them from general knowledge. When I added this to the resume prompts, agents produced honest "확인하지 못한 것 / not verified" sections that made the coverage boundary auditable. Without it, a truncated agent has an incentive to paper over gaps.

A + C together are cheap, need no threshold changes, and would likely have avoided this entirely. B is worth doing regardless.

## Note

The analysis quality, once recovered, was genuinely good — `seo-schema` caught two real defects in a theme patch I had drafted (a dead `og:image` URL returning 404, and the missing Google-required `image` field), and `seo-sxo` ran 7 live SERP checks to reach a well-evidenced targeting verdict. The agents work. They just aren't given enough runway to hand their work back.

Ilno:
Closes #177, closes #272.

## The failure

Both issues describe the same thing: an audit subagent finishes its research, gets terminated by the turn cap before reaching a write instruction placed at the end of its prompt, and the orchestrator sees no findings at all. #272 hit it on a **five-page Blogger site**, so this is not a large-site edge case — the write is simply the last thing in the prompt, and the last thing is what a cap removes.

## Changes

Two changes to the 15 agents that own a `findings/<category>.md` deliverable.

**1. Turn budgets clear the observed ceiling.**
#177 recorded stops at up to **32 tool calls** against caps of 15–25, so every audit agent could be killed before writing. Budgets are tripled, which preserves the relative ordering already encoded in 15/20/25 (how heavy each agent is) while clearing 32 with headroom:

| was | now | agents |
|---:|---:|---|
| 15 | 45 | content, google, performance, schema, sitemap, visual, drift |
| 20 | 60 | backlinks, cluster, ecommerce, geo, local, sxo, technical |
| 25 | 75 | maps |

**2. The write moves to the front and becomes incremental.**
A new `## Findings file — open it FIRST, append as you go` section sits directly after the frontmatter, ahead of every analysis instruction. Each agent creates its file with an `_in progress_` marker as its **first tool call**, appends each finding as it confirms it, and drops the marker when done.

That ordering is the actual fix. A file opened first leaves usable partial findings on disk when the cap hits; a file written last leaves nothing. It converts a total failure into a graceful one, which is the third suggestion in #177. The existing trailing section is left in place.

## One correction to #177

The issue reports that *"13 of 18 `seo-*` agents have no persistence instruction at all"*, listing `seo-geo`, `seo-local`, `seo-sxo` and `seo-sitemap` among them. That undercounts: those four **do** have one, under the heading `## Audit Persistence` rather than `## Persistence Contract`, so a grep for the latter misses them.

The real distribution is **15 of 18** with a findings contract. The three without — `seo-dataforseo`, `seo-flow`, `seo-image-gen` — are not audit agents and are never spawned by `/seo audit`, so they are correctly left untouched here.

So the defect is **ordering and turn budget, not absence**, and that is what this PR fixes. Flagging it because a PR written to the issue's description would have added dead instructions to three agents that do not participate in audits.

## Tests

Adds `tests/test_audit_persistence.py`, 47 tests:

- every audit agent's `maxTurns` exceeds the 32-turn overrun recorded in #177
- its write instruction precedes its `## Output Format` / `## Output Rules` section
- it has the `Write` tool
- every agent the audit skill delegates to owns a findings file — a drift guard between `skills/seo-audit/SKILL.md` and `agents/`
- a guard on the guard, so the parametrized suite cannot pass vacuously if agent discovery ever returns nothing

Confirmed these **fail on current `main`** (30 failures) before the fix, so they pin the behaviour rather than describe it.

## Verification

```
pytest tests/ -q
488 passed in 195.70s
```

441 before this PR, 488 after (+47), no regressions.

## Note

I have not been able to reproduce the original end-to-end run (no audit against a live site with subagent delegation on my side), so this is verified against the agent definitions and the recorded turn counts in the two issues, not against a fresh full audit. If you have a site that reproduced #272, that would be the real confirmation.


Ilno:
Closing this one, sorry for the noise. Opened it without the go-ahead on my side; the analysis is in the thread if it is useful to you, feel free to take any of it.

PR #292, opened 2026-09-07
fix(consistency): fold CRLF before hashing FLOW lock inputs

Nordalux:
## Summary

`consistency_check.py` hashes the 44 FLOW-locked prompt files byte-for-byte against `flow-prompts.lock`. The lock was written from LF content (`sync_flow.py` hashes the fetched text), but Git for Windows defaults to `core.autocrlf=true`, so a stock clone checks those files out with CRLF. Every locked file then reads as tampered: `consistency_check.py` reports 44 errors and `test_consistency_check` fails on any Windows machine cloned the way the README describes it.

Fold CRLF to LF before hashing. LF content hashes unchanged, so the lock still fails on any real edit; the second new test pins that.

```
$ git config core.autocrlf   # true, Git for Windows default
$ python scripts/consistency_check.py --json | jq '.errors | length'
44                            # before
0                             # after
```

Verified in:
- Windows 11, Python 3.14, clean venv from `requirements.txt` + pytest, `core.autocrlf=true` — full suite 431 passed, 13 skipped (the 13 are POSIX mode-bit tests, skipped by the follow-up CI PR)
- Windows 11, Python 3.11.16, same setup — 431 passed, 13 skipped
- GitHub Actions `windows-latest`, Python 3.10, full suite with the portability job widened (follow-up PR) — `test_no_consistency_errors` fails on `main` with `flow lock: hash mismatch` and passes with this commit (the runner's git also ships `autocrlf=true`)
- GitHub Actions `macos-latest` and `ubuntu-latest`, Python 3.10 — green

Both new tests fail without the change to `consistency_check.py` (3 failed, 2 passed).

Note: this and the follow-up CI PR both add an `[Unreleased]` section to `CHANGELOG.md`; whichever lands second needs a trivial rebase, which I am happy to do.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- `consistency_check.py` reported every FLOW-locked prompt file as a hash
+  mismatch on a stock Windows clone: Git for Windows checks the files out with
+  CRLF (`core.autocrlf=true`) while the lock hashes LF content. The check now
+  folds CRLF before hashing, so the lock only fails on real content changes.
+
 ## [2.2.5] - 2026-08-25
 
 ### Added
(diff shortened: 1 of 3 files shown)


AgriciDaniel:
Landed in v2.2.6 as commit 6c7bad (cherry-picked with -x, your authorship preserved). Thank you.

PR #294, opened 2026-09-07
fix(url_safety): refuse the RFC 6598 shared address space

Nordalux:
## Summary

`is_safe_ip` decides what counts as public by `ipaddress.is_private` plus a few flags. That leaves the RFC 6598 shared address space (100.64.0.0/10) open: Python does not classify it as private on any supported version, but it is never publicly routable. Alibaba Cloud serves its instance metadata at 100.100.100.200, and mesh VPNs hand out 100.64/10 addresses for internal services, so a URL literal or a DNS answer in that range passed `validate_url_strict` and the pinned-resolver check.

Refuse the range explicitly. IPv4-mapped IPv6 literals are unwrapped first, so every IPv4 rule also covers `::ffff:a.b.c.d`.

```
>>> validate_url_strict("http://100.100.100.200/latest/meta-data/")
('http://100.100.100.200/latest/meta-data/', '100.100.100.200')   # before
URLSafetyError: Blocked IP literal: 100.100.100.200                # after
```

Verified in:
- Windows 11, Python 3.14, clean venv from `requirements.txt` + pytest — `test_url_safety.py` 99 passed; full suite 433 passed, 17 failed — the same 17 that fail on `main` on Windows today (POSIX mode bits, cp1252 decode, CRLF lock), addressed by #292/#293
- Windows 11, Python 3.11.16, same setup — 432 passed, 18 failed — the same 18 that fail on `main` on Windows today (POSIX mode bits, cp1252 decode, CRLF lock, `os.fchmod` absent before 3.13), addressed by #292/#293
- GitHub Actions `ubuntu-latest`, Python 3.10, full suite — 450 passed

The added `is_safe_ip` cases fail without the change (3 failed on `main`).

Note: this and #292/#293 each add an `[Unreleased]` section to `CHANGELOG.md`; whichever lands later needs a trivial rebase, which I am happy to do.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- `url_safety.is_safe_ip` accepted the RFC 6598 shared address space
+  (100.64.0.0/10), which Python's `ipaddress` does not count as private. Alibaba
+  Cloud serves instance metadata at 100.100.100.200, so a crafted URL or DNS
+  answer in that range slipped past the SSRF guard. The range is now refused,
+  and IPv4-mapped IPv6 literals are judged as their embedded IPv4 address.
+
 ## [2.2.5] - 2026-08-25
 
 ### Added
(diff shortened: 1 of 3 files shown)


AgriciDaniel:
Landed in v2.2.6 as commit a63e33 (cherry-picked with -x, your authorship preserved). Thank you.

PR #295, opened 2026-09-07
fix(url_safety): let the configured HTTP proxy resolve inside the pinned scope

### LINKED ISSUE #280: url_safety.py DNS-pinning rejects the process's own HTTP proxy (127.0.0.1), breaking every live fetch behind an HTTP CONNECT proxy
## Summary

`scripts/url_safety.py`'s `safe_requests_get`/`safe_requests_head` fail for **every** target URL, including otherwise-reachable, in-policy ones, when the process is configured to route outbound HTTPS through a local HTTP CONNECT proxy (`HTTPS_PROXY=http://127.0.0.1:<port>`, a common sandbox/CI setup). This breaks `fetch_page.py`, `parse_html.py`, `sitemap_discovery.py`, `verify_backlinks.py`, `preload_check.py`, `parasite_risk.py`, `gbp_deprecation_lint.py`, `ucp_check.py`, `indexnow_submit.py`, and anything else built on the shared safe-fetch helpers, roughly 18 scripts.

## Root cause

`_pin_dns()`'s fallthrough validation (the "Branch 2" comment: *"every OTHER hostname (redirect target, embedded subresource, library bookkeeping) gets resolved by the real resolver, then each returned record is checked"*) runs `is_safe_ip()` against **any** hostname resolved during the pinned scope, not just ones actually related to the target site.

When `requests`/`urllib3` is configured with an HTTP proxy, it must resolve the proxy's own host to open the CONNECT tunnel. If that proxy is local (`127.0.0.1`, as in many container/sandbox setups), `socket.getaddrinfo('127.0.0.1', ...)` runs inside the patched scope, hits Branch 2, and `is_safe_ip('127.0.0.1')` correctly reports it as loopback/non-public — so the wrapper raises `socket.gaierror`, refusing the library's own connection to the proxy it needs in order to reach the internet at all.

## Repro

```python
import sys
sys.path.insert(0, 'scripts')
import url_safety

# Any HTTPS_PROXY pointing at localhost reproduces this for any target:
r = url_safety.safe_requests_get('https://example.com/', timeout=15)
```

```
requests.exceptions.ProxyError: HTTPSConnectionPool(host='example.com', port=443):
Max retries exceeded with url: / (Caused by ProxyError('Unable to connect to proxy',
NameResolutionError("HTTPSConnection(host='127.0.0.1', port=<port>): Failed to resolve
'127.0.0.1' ([Errno -4] url_safety: refused to resolve '127.0.0.1' to non-public IP
127.0.0.1)")))
```

Confirmed target-independent: reproduces identically against both a fully in-policy site (`https://trumould.com/sitemap.xml`, which returns 200 via a plain `curl` through the same proxy) and an org-blocked one (`https://api.indexnow.org/`) — same exception, same `127.0.0.1` resolution refusal, in both cases before any request reaches the actual target or even the proxy's CONNECT step.

Downstream effect observed in a `claude-seo` skill run: `sitemap_discovery.py https://trumould.com --json` returns `"error": "Request failed"` for every candidate URL with no further detail, which reads as "target unreachable" when the actual cause is entirely local and unrelated to the target.

## Suggested fix

The DNS-pinning/anti-rebinding check exists to stop a *target* hostname's redirect chain or subresources from rebinding to an internal address — it should not apply to the transport's own proxy connection, which is operator-configured infrastructure, not attacker-influenced input. A few options:
- Exempt the configured proxy host (from `requests`' resolved proxy settings, or `HTTPS_PROXY`/`HTTP_PROXY`/`ALL_PROXY` env vars) from the Branch 2 fallthrough check.
- Or scope the patched `getaddrinfo` to only intercept lookups for the pinned target hostname and known-redirect hostnames actually encountered via `Response.history`, rather than intercepting *every* `getaddrinfo` call process-wide during the request.
- At minimum, surface a distinguishable error (e.g. a specific exception subclass or message) when the refusal is against the proxy itself, so callers like `sitemap_discovery.py` can report "toolkit fetch layer misconfigured for this environment" instead of a generic "Request failed" that looks identical to a genuinely unreachable target.

## Environment

Reproduced in a sandboxed Claude Code session with outbound HTTPS routed through a local HTTP CONNECT proxy at `127.0.0.1:<port>` (via `HTTPS_PROXY`/`https_proxy`). This is a common pattern for sandboxed/CI agent environments, so the fix likely matters beyond this one setup.

Nordalux:
## Summary

Fixes #280. With `HTTPS_PROXY=http://127.0.0.1:<port>` (the usual sandbox and CI setup) every `safe_requests_get` / `safe_requests_head` / `safe_requests_session` call failed before anything left the process: `requests` has to resolve the proxy's own host to open the CONNECT tunnel, that lookup ran through the pinned resolver's fall-through validation, and `127.0.0.1` was refused as non-public. Reproduced on `main` against `https://example.com/`:

```
ProxyError(... NameResolutionError("... Failed to resolve '127.0.0.1'
  ([Errno 11003] url_safety: refused to resolve '127.0.0.1' to non-public IP 127.0.0.1)"))
```

The change computes the proxy host `requests` will select for the URL (environment proxies honouring `NO_PROXY`, overridden by an explicit `proxies=` mapping) and exempts exactly that host from the resolution check for the duration of the fetch. Redirect targets and every other hostname are still validated; a test pins that the exemption does not leak. Behind a CONNECT proxy the target is resolved by the proxy, so `validate_url_strict`'s pre-flight check remains the guard there, as the docstring now says.

Verified in:
- Windows 11, Python 3.14, clean venv from `requirements.txt` + pytest — `test_url_safety.py` 96 passed; full suite 430 passed, 17 failed — the same 17 that fail on `main` on Windows today (POSIX mode bits, cp1252 decode, CRLF lock), addressed by #292/#293
- Same machine, a minimal HTTP CONNECT proxy on `127.0.0.1` with `HTTPS_PROXY` pointed at it: `safe_requests_get("https://example.com/")` returns 200 through the tunnel on this branch; on `main` the identical script stops at `url_safety: refused to resolve '127.0.0.1'`
- Windows 11, Python 3.11.16, same setup — 429 passed, 18 failed — the same 18 that fail on `main` on Windows today (POSIX mode bits, cp1252 decode, CRLF lock, `os.fchmod` absent before 3.13), addressed by #292/#293
- GitHub Actions `ubuntu-latest`, Python 3.10, full suite — 447 passed

The six new tests are offline (resolver and proxy table are patched) and fail without the change (5 failed on `main`; the sixth needs the new keyword).

Note: this and #292/#293 each add an `[Unreleased]` section to `CHANGELOG.md`; whichever lands later needs a trivial rebase, which I am happy to do.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- Every live fetch failed behind an HTTP proxy on loopback
+  (`HTTPS_PROXY=http://127.0.0.1:<port>`, the usual sandbox and CI setup): the
+  pinned resolver refused to resolve the proxy's own address as non-public. The
+  proxy host `requests` selects for the URL is now exempt from that check, and
+  only that host. (#280)
+
 ## [2.2.5] - 2026-08-25
 
 ### Added
--- a/scripts/url_safety.py
+++ b/scripts/url_safety.py
@@ -345,6 +345,24 @@ def validate_url_strict(url: str) -> tuple[str, str]:
     return url, pinned
 
 
+def _proxy_hosts(url: str, proxies: Optional[dict] = None) -> frozenset:
+    """Hostnames ``requests`` will connect to instead of ``url``'s host.
+
+    Mirrors the proxy selection ``requests`` performs for a plain call:
+    environment proxies (``HTTPS_PROXY`` and friends, honouring ``NO_PROXY``)
+    overridden by an explicit ``proxies`` mapping. Returns the empty set when
+    the request goes direct.
+    """
+    merged = dict(requests.utils.get_environ_proxies(url))
+    if proxies:
+        merged.update(proxies)
+    proxy = requests.utils.select_proxy(url, merged)
+    if not proxy:
+        return frozenset()
+    host = urlparse(requests.utils.prepend_scheme_if_needed(proxy, "http")).hostname
+    return frozenset({host.lower()}) if host else frozenset()
+
+
 # A single non-blocking lock guards the global getaddrinfo monkey-patch.
 # This is a deliberate choice: claude-seo scripts run one URL fetch at a
 # time, and we'd rather raise loudly than silently corrupt resolver state
@@ -353,7 +371,12 @@ _dns_patch_lock = threading.Lock()
 
 
 @contextmanager
-def _pin_dns(hostname: str, pinned_ip: str, port: int) -> Iterator[None]:
+def _pin_dns(
+    hostname: str,
+    pinned_ip: str,
+    port: int,
+    exempt_hosts: frozenset = frozenset(),
+) -> Iterator[None]:
     """
     Temporarily override ``socket.getaddrinfo`` so the named host resolves
     only to ``pinned_ip``, AND every other hostname looked up during the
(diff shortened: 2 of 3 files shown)


AgriciDaniel:
The exemption is correctly narrow for redirect targets (verified: a target resolving to 192.168.1.10 is still refused inside the pinned scope), and the tests are good. One gap: the proxy host itself never goes through the blocklist. With HTTPS_PROXY pointing at a loopback listener at this head, main raises ProxyError and this branch returns the proxy's body, so anything that can set the proxy variable gets unchecked egress to an internal host. Please run the resolved proxy host through the same blocklist before adding it to the exempt set, and add a negative control for a metadata address in HTTPS_PROXY. One sentence in SECURITY.md would also help: behind a CONNECT proxy, pinning degrades to a pre-flight check.

AgriciDaniel:
Landed in v2.3.0 as commit d601422; the proxy host is now validated before exemption and re-checked at resolve time. Thank you.

PR #296, opened 2026-09-07
fix(preload_check): exit 0 on a completed run, gate only with --fail-under

### LINKED ISSUE #281: preload_check.py exits 1 on a successful run (exit code encodes score, not status)
# preload_check.py exits 1 on a successful run (exit code encodes score, not status)

**Operating System:** Windows 11 Home Single Language 10.0.26200 (manual install, v2.2.5)

**Python Version:** 3.12.10

**Command that failed:**

```
./bin/claude-seo run preload_check.py https://example.com --json
```

**Full error output:**

There is no error output. The script emits complete, well-formed JSON on stdout and then exits 1:

```json
{
  "url": "https://example.com/",
  "speculation_rules": { "inline_blocks": 0, "header_present": false, "actions": [] },
  "preload_hints": 0,
  "prerender_links": 0,
  "bfcache_signals": { "cache_control_no_store": false, "unload_listener": false, "beforeunload_listener": false },
  "lcp_resource_hints": { "preload_lcp_candidate": false, "fetchpriority_high": 0 },
  "score": 50,
  "recommendations": [ ... ]
}
```

```
$ ./bin/claude-seo run preload_check.py https://example.com --json >/dev/null 2>&1; echo $?
1
```

**Expected behavior:**

Exit 0. The analysis succeeded — the URL was fetched, parsed, scored, and serialized without error. A low score is a *finding*, not a failure of the run.

---

## Root cause

`scripts/preload_check.py:232`:

```python
return 0 if result["score"] >= 75 else 1
```

The exit code encodes a quality threshold rather than execution status. Because `example.com` scores 50, a fully successful run reports failure.

## Why this matters

1. **It collides with the genuine error path.** Line 205 returns `2` for `URLSafetyError`. So `1` means "ran fine, scored below 75" while `2` means "actually failed" — but every generic caller treats *any* non-zero as failure. A `set -e` script, a CI step, or a wrapper checking `if result.ExitCode -ne 0` will abort on a healthy run.

2. **The threshold is hardcoded and undocumented.** `75` is a magic number with no flag to change it, and the module docstring (lines 1–25) documents what the script detects but says nothing about exit-code semantics.

3. **It is inconsistent with the one other script that does this.** `scripts/content_quality.py:287` uses the same pattern but makes the threshold configurable:

   ```python
   return 0 if result["overall_quality"] >= args.threshold else 1
   ```

   `preload_check.py` offers no equivalent `--threshold`, so callers cannot opt out of the gate.

These are the only two scripts of 53 that encode a score in the exit status, so this reads as an unintended divergence rather than a project-wide convention.

## Suggested fix

Any one of these would resolve it; listed in order of preference:

1. **Return 0 whenever the analysis completes.** Callers that want a gate can read `score` from the JSON. Cleanest, and matches the other 51 scripts.
2. **Add an opt-in `--fail-under N` flag**, defaulting to no gating. Preserves the CI use case without breaking the default path.
3. **At minimum, document it** — add exit-code semantics to the module docstring and `docs/COMMANDS.md`, and mirror `content_quality.py`'s `--threshold` so the value isn't hardcoded.

## Notes

Reproduced on Windows, but the code path is platform-independent (pure `argparse` + return value), so this should reproduce anywhere. Found while smoke-testing a fresh v2.2.5 install; the analysis output itself was correct and useful — only the exit status is wrong.

Nordalux:
## Summary

Fixes #281. `preload_check.py` encoded a quality threshold in its exit status: any page scoring below a hard-coded 75 made a successful run exit 1, so `set -e` scripts and CI steps aborted on healthy pages, and 1 was indistinguishable in meaning from the genuine error path (2 for a URL refused by url_safety). The threshold was undocumented and not configurable, unlike `content_quality.py`'s `--threshold`.

A completed analysis now exits 0. The gate stays available as an opt-in `--fail-under N`, and the module docstring documents the exit status (0 / 1 / 2).

```
$ python scripts/preload_check.py https://example.com/ --json >/dev/null; echo $?
1     # main (score 50)
0     # this branch
$ python scripts/preload_check.py https://example.com/ --json --fail-under 75 >/dev/null; echo $?
1     # opt-in gate, same behaviour as before
```

Verified in:
- Windows 11, Python 3.14, clean venv from `requirements.txt` + pytest — `test_technical_depth.py` 20 passed; full suite 427 passed, 17 failed — the same 17 that fail on `main` on Windows today (POSIX mode bits, cp1252 decode, CRLF lock), addressed by #292/#293; the three CLI runs above against the live `https://example.com/`
- Windows 11, Python 3.11.16, same setup — 426 passed, 18 failed — the same 18 that fail on `main` on Windows today (the 17 above plus `os.fchmod` absent before 3.13), addressed by #292/#293
- GitHub Actions `ubuntu-latest`, Python 3.10, full suite — 444 passed

Two of the three new tests fail without the change (the third pins that exit 2 is unchanged).

Note: this and #292–#295 each add an `[Unreleased]` section to `CHANGELOG.md`; whichever lands later needs a trivial rebase, which I am happy to do.

## Type of Change

- [x] Bug fix
- [ ] New feature / sub-skill
- [ ] Documentation update
- [ ] Refactor / code quality
- [ ] Other (describe below)

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- `preload_check.py` exited 1 on every successful run that scored below a
+  hard-coded 75, so `set -e` scripts and CI steps aborted on healthy pages. A
+  completed analysis now exits 0; the gate is opt-in through `--fail-under N`.
+  Exit 2 for a URL refused by url_safety is unchanged. (#281)
+
 ## [2.2.5] - 2026-08-25
 
 ### Added
--- a/scripts/preload_check.py
+++ b/scripts/preload_check.py
@@ -46,6 +46,12 @@ Result JSON::
       "score": 0..100,
       "recommendations": ["…"]
     }
+
+Exit status::
+
+    0  analysis completed (a low score is a finding, not a failed run)
+    1  score below ``--fail-under N`` (opt-in gate, off by default)
+    2  URL refused by url_safety
 """
 
 from __future__ import annotations
@@ -195,6 +201,14 @@ def main() -> int:
     )
     parser.add_argument("url")
     parser.add_argument("--json", action="store_true")
+    parser.add_argument(
+        "--fail-under",
+        type=int,
+        default=None,
+        metavar="N",
+        help="Exit 1 if the score is below N. Off by default: a successful "
+             "run exits 0 whatever the score.",
+    )
     args = parser.parse_args()
 
     try:
@@ -228,7 +242,9 @@ def main() -> int:
             for r in result["recommendations"]:
                 print(f"  - {r}")
 
-    return 0 if result["score"] >= 75 else 1
+    if args.fail_under is not None and result["score"] < args.fail_under:
+        return 1
+    return 0
 
 
 if __name__ == "__main__":
--- a/tests/test_technical_depth.py
+++ b/tests/test_technical_depth.py
@@ -14,8 +14,10 @@ deferred to manual smoke tests.
 
 from __future__ import annotations
 
+import json
 import os
 import sys
+from types import SimpleNamespace
 from unittest.mock import patch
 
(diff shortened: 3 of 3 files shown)


AgriciDaniel:
Landed in v2.2.6 as commit b43dd9 (cherry-picked with -x, your authorship preserved). Thank you.

PR #297, opened 2026-09-09
feat(fetch-page): add JSON output for raw and rendered fetches

### LINKED ISSUE #282: fetch_page.py has no --json flag, unlike the other 42 scripts that do
# fetch_page.py has no --json flag, unlike the other 42 scripts that do

**Operating System:** Windows 11 Home Single Language 10.0.26200 (manual install, v2.2.5)

**Python Version:** 3.12.10

**Command that failed:**

```
./bin/claude-seo run fetch_page.py https://example.com --json
```

**Full error output:**

```
usage: fetch_page.py [-h] [--output OUTPUT] [--timeout TIMEOUT]
                     [--no-redirects] [--user-agent USER_AGENT]
                     [--render {auto,always,never}]
                     url
fetch_page.py: error: unrecognized arguments: --json
```

Exit code: `2`

**Expected behavior:**

Either `--json` is accepted and emits structured output like the rest of the toolset, or the docs make clear that `fetch_page.py` is text-only so callers don't reach for a flag that doesn't exist.

---

## Context

42 of the 53 scripts in `scripts/` accept `--json`. `fetch_page.py` is one of the few that doesn't, and it's among the most likely to be called first in an audit chain, since it's the basic retrieval primitive.

The inconsistency is easy to trip over because `SKILL.md` and the sub-agent instructions list `fetch_page.py` alongside `render_page.py`, `parse_html.py`, and `sitemap_discovery.py` as interchangeable data sources — all three of those *do* support `--json`. Nothing signals that one member of that group behaves differently.

Its current output is a human-readable header plus the raw body:

```
URL: https://example.com/
Status: 200
<!doctype html>...
```

Parsing that programmatically means string-splitting on `URL:` / `Status:`, which is brittle. It also discards data the script already has in hand — response headers and the redirect chain are fetched but never surfaced, whereas `render_page.py` returns both.

## Suggested fix

Add `--json` emitting at minimum:

```json
{
  "url": "https://example.com/",
  "status_code": 200,
  "content": "...",
  "headers": { },
  "redirect_chain": []
}
```

Reusing `render_page.py`'s field names would let callers switch between the two without remapping keys.

If text-only output is intentional — reasonable, given the script's job is to hand back a body — then the lighter fix is a note in `docs/COMMANDS.md` and in the sub-agent tool lists stating that `fetch_page.py` does not support `--json` and pointing at `render_page.py` for structured retrieval.

## Notes

Filing this as a consistency/DX report rather than a defect: the script works correctly at what it does, and `argparse` rejecting an undefined flag is proper behavior. The problem is that the toolset advertises a uniform interface it doesn't quite have. Found while smoke-testing a fresh v2.2.5 install, where two separate agents independently reached for `--json` on this script.

Formatted:
Fixes #282.

`fetch_page.py URL --json` currently exits 2 because the flag is unsupported. Add JSON output for raw HTTP and both render modes, exposing the complete result already collected by the fetcher: content, response headers, redirect metadata, and errors.

`--json --output page.html` saves successful HTML separately and emits only JSON to stdout, with `output_written` indicating whether the file was saved. Fetch errors emit JSON and exit 1 without overwriting an existing output file. Existing text output and HTTP-status behavior are preserved. Includes command documentation and a changelog entry.

Validation on macOS, Python 3.12:
- Full suite: 459 passed.
- Focused fetch/renderer suites: 72 passed, including 18 new CLI regression cases.
- Ruff on changed Python files and `git diff --check`: passed.
- Managed-runtime portability check: 33 skills checked, zero errors or warnings.
- Live example.com fetches through the managed launcher in `never`, `auto`, and `always` modes: HTTP 200, parseable JSON, and saved HTML matching `content`; `always` used Chromium.
- Blocked loopback URL: parseable JSON error and exit 1.


AgriciDaniel:
The flag is wanted (#282). The rendered path builds its own JSON instead of reusing render_page._json_summary, so the two contracts diverge and the --max-text bound added in v2.2.5 is lost on this path. Please route the rendered branch through the existing summary function and add a test asserting the same keys on both paths.

AgriciDaniel:
Landed in v2.3.0 as commit e7fdd9b; the raw and rendered paths now share one JSON contract with --max-text on both. Thank you.

PR #300, opened 2026-09-10
fix(runtime): surface the failing setup stage's own output

Nordalux:
`_run_checked` discarded the child's stderr, so a failed venv or pip stage reported only `failed with exit code 1`. `venv` additionally swallows `ensurepip`'s output, so a pip bootstrap failure was invisible one level deeper.

pip is now bootstrapped as its own stage, the last lines of the failing stage's output are appended to the error, and the home-directory redaction matches the repr-quoted and mixed-case forms that child tracebacks print.

Before/after on a checkout whose data dir exceeds MAX_PATH:

```
Claude SEO setup failed: virtual environment creation failed with exit code 1
---
Claude SEO setup failed: pip bootstrap failed with exit code 1
  ERROR: Could not install packages due to an OSError: [WinError 206] Der Dateiname oder die Erweiterung ist zu lang: '<home>\\...\\.venv.next-81952\\Lib\\site-packages\\pip-26.0.1.dist-info\\licenses\\src\\pip\\_vendor\\dependency_groups'
```

Verified in:
- Windows 11, Python 3.14.4, repo checkout — full suite 512 passed, 17 skipped; `install.sh` v2.2.6 into a fresh short-path HOME still completes and `doctor` reports ready
- Windows 11, Python 3.11.16 (uv, pytest+requests+lxml only) — test_runtime.py and test_runtime_installers.py 19 passed, 1 skipped
- Fork CI on Nordalux/claude-seo @1f953b4 — test suite, portability (windows-latest, macos-latest), dependency audit, syntax check, secret scan all green

### DIFF
--- a/scripts/runtime.py
+++ b/scripts/runtime.py
@@ -198,11 +198,22 @@ def _safe_env(status: dict[str, Any]) -> dict[str, str]:
     return env
 
 
+def _home_pattern(home: str) -> re.Pattern[str]:
+    # Child output quotes paths in repr form (doubled or quadrupled separators
+    # once nested) and Windows tools mix drive-letter case, so match the home
+    # directory by its segments rather than by the exact string.
+    parts = [re.escape(part) for part in re.split(r"[\\/]+", home) if part]
+    pattern = r"[\\/]+".join(parts)
+    if home[:1] in ("\\", "/"):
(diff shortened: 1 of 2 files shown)


AgriciDaniel:
AgriciDaniel/claude-seo

PR #315, opened 2026-09-16
Fix SEO audit agent preflight

morus246:
## Summary

- add an audit-agent health checker for the required full-audit subagents
- surface audit subagent availability in `claude-seo doctor` across source, plugin, project-local, and manual layouts
- require `/seo audit` and the main `/seo` orchestrator to disclose inline fallback instead of claiming the full subagent pipeline when agents are missing or broken

## Validation

- `PYTEST_ADDOPTS="-p no:cacheprovider" python3 -m pytest tests/ -q` -> `776 passed, 3 skipped`
- `bash scripts/claude-seo doctor` in the source worktree prints `Audit subagent check: ok`; it exits 3 only because the managed runtime is not set up there

## Notes

- fixes broken-agent/link trust issues found while validating `/seo audit` usage
- no private Fabiomorus/local data is included in the diff


AgriciDaniel (changes requested):
Thanks, this is useful and read-only. Two requests before merge:
1. Add `seo-agentic` to the required-agent list. v2.4.0 makes it a standard audit agent, so doctor would report "ok" while it is missing.
2. Import the new module inside the doctor command (or fail soft). A top-level import in runtime.py means a packaging miss would break every runtime command, not only doctor.

Expect a small text conflict in skills/seo/SKILL.md after rebasing on v2.4.0.

PR #318, opened 2026-09-21
feat(security): allow reserved .test targets safely

joachimBrindeau:
## Summary

- allow the explicit reserved-TLD form `*.test[:port]` in `CLAUDE_SEO_LOCAL_TARGETS`
- preserve exact `host[:port]` and IPv6 entries with strict fail-closed parsing
- bind pinned DNS to hostname and effective port so redirects cannot cross ports
- prevent operator exemptions from reaching WHOIS referrals, Unlighthouse, robots-declared sitemaps, or UCP capability endpoints
- document the top-level-only capability and add adversarial regression coverage

Generic wildcards remain unsupported. Browser subresources and redirects still do not inherit local-target exemptions.

## Verification

- `python3 -m pytest tests -q` — 772 passed, 3 skipped
- `python3 scripts/consistency_check.py` — 0 errors, 0 warnings
- `claude-seo run portability_check.py` — 0 errors, 0 warnings
- Python compile checks passed for all modified scripts
- `git diff --check` passed
- live raw fetch with `CLAUDE_SEO_LOCAL_TARGETS=*.test`: `https://klarc.test/` returned HTTP 200 and 323,639 bytes
- independent adversarial security review: APPROVE


AgriciDaniel (changes requested):
Thank you for the stricter parser and for binding the DNS pin to the validated port; we would like to keep both. Three things block merge:
1. The second commit (33b25392) rewrites four skills with instructions from your own workspace (Umami, Hermes CLI, files that do not exist here). It looks accidental, but for every user it would act as instructions to their agent. Please drop it.
2. `allow_local_target` defaults to True, so callers that fetch discovered third-party URLs still get the exemption. Please default it to False and opt in only where the target is operator-named.
3. RFC 6761 section 6.2 says application software "SHOULD NOT recognize test names as special", and SECURITY.md promises no wildcard. If an exemption stays, please accept only explicit `*.test:<port>` entries.

